PR #94, opened 2018-12-02
Polishing app metadata

jancborchardt:
- Add description for app management, fix navigation sorting
- Adjust readme text
- Remove spacer element in navigation, not needed at the moment (this was hover/focusable for example, non-standard)
- Add section on current limitations to readme, as you mentioned via mail @daita 

Please review @juliushaertl @daita


m4sk1n:
So, is it going to implement protocol used by Diaspora?

jancborchardt:
@m4sk1n likely not, but Diaspora is working to implement ActivityPub.

Issue #153, opened 2018-12-07
[0.1.x] Known issues.


daita:
- [ ] Creating a post as 'Unlisted' is not working as expected.
- [ ] message saying that '.well-known is wrongly configured' is displayed on the first launch of the app. A simple refresh fix this.  @juliushaertl 
- [ ] missing Notification Stream (so if anyone is taggin you on a public message, you wont see the message ...)
- [ ] You don't see your own posts in the Home Stream if you're following no one.  → https://github.com/nextcloud/social/issues/166

jancborchardt:
@daita (cc @juliushaertl) same here, please open individual issues and use the Milestone to manage. :)

PR #169, opened 2018-12-07
Run well-known check after setup

juliushaertl:
Fixes  message saying that '.well-known is wrongly configured' is displayed on the first launch of the app. A simple refresh fix this. 

from #153


jancborchardt:
@juliushaertl @daita @MorrisJobke is Drone configured & running for this repository btw? Enabled it in the pull request while I was there because it seemed like it was enabled.

PR #176, opened 2018-12-08
Cleaning some TODO

daita:
Adding the CSRF check on internal API request.
Fixing the generation of unique Id on Follow/Accept

### DIFF
--- a/lib/Controller/NavigationController.php
+++ b/lib/Controller/NavigationController.php
@@ -271,10 +271,6 @@ class NavigationController extends Controller {
 
 
 	/**
-	 *
-	 * // TODO: Delete the NoCSRF check
-	 *
-	 * @NoCSRFRequired
 	 * @NoAdminRequired
 	 * @NoSubAdminRequired
 	 *
@@ -295,9 +291,6 @@ class NavigationController extends Controller {
 
 
 	/**
-	 *
-	 * // TODO: Delete the NoCSRF check
-	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
 	 * @NoAdminRequired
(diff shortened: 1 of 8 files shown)


lib/Controller/NavigationController.php
```diff
@@ -295,9 +291,6 @@ public function documentGet(string $id): Response {
 
 
 	/**
-	 *
-	 * // TODO: Delete the NoCSRF check
-	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
```
nickvergessen: forgot to remove?

lib/Controller/QueueController.php
```diff
@@ -83,8 +83,6 @@ public function __construct(
 
 
 	/**
-	 * // TODO: Delete the NoCSRF check
-	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
 	 * @NoAdminRequired
```
nickvergessen: PublicPage makes No(Sub)AdminRequired obsolete

daita: As said, I will make some more test after every PRs are merged regarding the CSRF.
Here, I need to check if doing local curl request will work without the @NoCSRFRequired.

I will also clean the No(Sub)AdminRequired in another PR

lib/Controller/NavigationController.php
juliushaertl: No, it is required here, since this is an endpoint users can navigate to.

lib/Controller/QueueController.php
juliushaertl: From where is this method requested?

daita: The QueueController is used to async process. The request is done from the app itself, locally.

daita:
fixing some obsolete No(Sub)AdminRequired.

PR #177, opened 2018-12-08
Async the process only if there is request in STANDBY

daita:


lib/Controller/QueueController.php
```diff
@@ -93,19 +93,21 @@ public function __construct(
...
 		$requests = $this->queueService->getRequestFromToken($token, RequestQueue::STATUS_STANDBY);
-		$this->activityService->manageInit();
-		foreach ($requests as $request) {
-			$request->setTimeout(ActivityService::TIMEOUT_ASYNC);
-			try {
-				$this->activityService->manageRequest($request);
-			} catch (RequestException $e) {
-			} catch (SocialAppConfigException $e) {
+
+		if (sizeof($requests) > 0) {
```
nickvergessen: `if (!empty($requests)) {`

You dont need to count the number, you just want to know if its empty or not

Issue #168, opened 2018-12-07
"Failed to create a post"


jancborchardt:
Not sure if this is about the characters used, or if only my first post after each page refresh is taken …:

Before posting
![social before posting](https://user-images.githubusercontent.com/925062/49677555-b3da0300-fa7f-11e8-860f-70c9c2afe072.png)

On posting I get a "Failed to create post" error and these errors show:
![social after posting](https://user-images.githubusercontent.com/925062/49677554-b3da0300-fa7f-11e8-9ac7-844ff12b976f.png)

@daita @juliushaertl

daita:
you had anything in the nextcloud.log ?

Kedoia:
same error for me, on nextcloud.log : 

`{"reqId":"6frGW87PJt5vR41maa92","level":2,"time":"2018-12-08T12:49:29+00:00","remoteAddr":"217.70.181.1","user":"valere","app":"no app in context","method":"POST","url":"\/apps\/social\/api\/v1\/post","message":"500 - {\"status\":-1,\"exception\":\"Doctrine\\\\DBAL\\\\Exception\\\\DriverException\",\"message\":\"An exception occurred while executing 'SELECT lastval()':\\n\\nSQLSTATE[55000]: Object not in prerequisite state: 7 ERREUR:  la derni\\u00e8re valeur (lastval) n'est pas encore d\\u00e9finie dans cette session\"}","userAgent":"Mozilla\/5.0 (X11; Ubuntu; Linux x86_64; rv:63.0) Gecko\/20100101 Firefox\/63.0","version":"15.0.0.9"}

`

daita:
#185 might fix your issue, @Kedoia but I am still curious about your nextcloud logs, @jancborchardt

daita:
should be fixed in 0.1.1

PR #185, opened 2018-12-08
removing useless getLastInsertId()

daita:
removing useless getLastInsertId() that seems to break on sqlite.


daita:
Should fix #168

PR #194, opened 2018-12-10
Bigger field to store data source for notes and actors cache

daita:
Using the migration, we set the VARCHAR(10000) for 'source' field in 2 tables.
We also confirm that the JSON is not too big to fit, in case it is, we save it as 'too_big'


lib/Db/CacheActorsRequest.php
```diff
@@ -66,6 +66,11 @@ public function __construct(
 	 * @return int
 	 */
 	public function save(Person $actor): int {
+		$source = $actor->getSource();
+		if (sizeof($source) >= CoreRequestBuilder::SOURCE_LENGTH) {
```
tcitworld: I wonder if this can bring issues with MySQL's UTF8 vs UTF8mb4 ?

nickvergessen: sizeof is a wrapper for count which is an array function anyway, so use count instead of sizeof.

Further more you should use `strlen($source)` or `isset($source[CoreRequestBuilder::SOURCE_LENGTH])` instead.

nickvergessen: @tcitworld well, you can add unit tests to check this ;)

daita: @nickvergessen not the first time, don't know why I keep using sizeof()

PR #196, opened 2018-12-10
Some data are only for internal use.

daita:
There is no need to send follow_id and accepted that are locally generated data when communicating to remote instance of ActivityPub

### DIFF
--- a/lib/Model/ActivityPub/Follow.php
+++ b/lib/Model/ActivityPub/Follow.php
@@ -128,10 +128,7 @@ class Follow extends ACore implements JsonSerializable {
 	 * @return array
 	 */
 	public function jsonSerialize(): array {
-		$result = array_merge(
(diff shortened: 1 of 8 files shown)


lib/Model/ActivityPub/Follow.php
```diff
@@ -128,13 +128,22 @@ public function importFromDatabase(array $data) {
 	 * @return array
 	 */
 	public function jsonSerialize(): array {
-		return array_merge(
+		$result = array_merge(
```
nickvergessen: array_merge with an empty array?

juliushaertl:
@daita Fine by me if the small comment by @nickvergessen is fixed.

daita:
I have also remove some others array_merge with empty arrays. This was just in prevision of more complex generated ld-json objects

Issue #199, opened 2018-12-10
Issue with post sent to a follower from a 3rd instances


daita:
**Describe the bug*
Cannot parse some post to followers.

**To Reproduce**
1. UserA follows an account (UserB) on another instance of ActivityPub
2. Post from UserC on a 3rd instance of ActivityPub to UserB
3. When incoming to the instance of UserA, the Post fail to be parsed because of the check on the Origin.

Origin is based on the instance who request the Create/Post (UserB->UserA), the Post itself is from another instance (UserC).

**Expected behavior**
Working.

PR #206, opened 2018-12-11
linked data signature

daita:
#199 - In case of a remote instance just bounce an object create from a 3rd instance.

We're using https://w3c-dvcg.github.io/ld-signatures to verify and sign objects so we can check their content

If available, the origin of the signed content is use to check the validity of the content. If not available, we keep using the origin from the header of the request.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
     }
   ],
   "require": {
-    "daita/my-small-php-tools": "dev-master"
+    "daita/my-small-php-tools": "dev-master",
+    "digitalbazaar/json-ld": "0.4.7"
   },
   "require-dev": {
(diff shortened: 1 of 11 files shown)


daita:
To test this PR, a `make composer` is mandatory !

jancborchardt:
> To test this PR, a `make composer` is mandatory !

Will this be mandatory for everything from this point on? If so, please add it to the readme. :)

daita:
This is now done. The signing and verifying process have been tested this way:

- UserA (Social App) follows UserB (Mastodon)
- UserC (Social App) follows UserB.
- UserC reply to a post from UserB and sign it.
- UserA receive the post and verify the origin.

juliushaertl:
Rebased to master.

PR #216, opened 2018-12-12
include the getAccount within the try/catch

daita:
It seems that getAccount() can returns an exception as PROPERTY_AVATAR might have a missing key ('verified')

### DIFF
--- a/lib/Service/ActorService.php
+++ b/lib/Service/ActorService.php
@@ -294,6 +294,9 @@ class ActorService {
 				$actor->setName($displayNameProperty->getValue());
 			}
 		} catch (Exception $e) {
+			$this->miscService->log(
+				'Issue while trying to updateCacheLocalActorName: ' . $e->getMessage(), 1
+			);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


lib/Service/ActorService.php
```diff
@@ -287,14 +286,14 @@ public function cacheLocalActorByUsername(string $username, bool $refresh = fals
...
-		$account = $this->accountManager->getAccount($user);
 
 		try {
+			$account = $this->accountManager->getAccount($user);
 			$displayNameProperty = $account->getProperty(IAccountManager::PROPERTY_DISPLAYNAME);
 			if ($displayNameProperty->getScope() === IAccountManager::VISIBILITY_PUBLIC) {
 				$actor->setName($displayNameProperty->getValue());
 			}
-		} catch (PropertyDoesNotExistException $e) {
+		} catch (Exception $e) {
```
tobiasKaminsky: I do not know when this function is used, but maybe some feedback that update failed is nice?

daita: The function is needed when we refresh the cache for local social accounts. It check the displayname in the settings of the nextcloud account. 

I have an issue with NC15 that I will investigate, but you're right I will add some logs in here

PR #217, opened 2018-12-12
varchar->text

### LINKED ISSUE #207: SQL Error: Row size too large when enabling Social app.
**Describe the bug**
Attempting to enable the Social app results in a failure with the following output:

An exception occurred while executing 'ALTER TABLE oc_social_cache_actors CHANGE `source` source VARCHAR(10000) NOT NULL COLLATE utf8mb4_bin': SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs

**To Reproduce**
Update Social to latest version from git.

**Expected behavior**
Social app gets enabled and begins being functional.

**Social app version:** 0.1 git

**Operating system**: Ubuntu 18.04.1

**Web server:** Apache 2.4.x

**Database:** MySQL 5.7.24

**PHP version:** 7.2 FPM

**Nextcloud version:** 15.0

daita:
Should fix #207 by replacing varchar with text in DB

### DIFF
--- a/lib/Migration/Version0001Date20181211222142.php
+++ b/lib/Migration/Version0001Date20181211222142.php
@@ -32,6 +32,9 @@ namespace OCA\Social\Migration;
 
 
 use Closure;
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Schema\SchemaException;
+use Doctrine\DBAL\Types\Type;
 use OCA\Social\Db\CoreRequestBuilder;
 use OCP\DB\ISchemaWrapper;
 use OCP\Migration\IOutput;
@@ -39,11 +42,11 @@ use OCP\Migration\SimpleMigrationStep;
 
 
 /**
- * Class Version0001Date20181208185242
+ * Class Version0001Date20181211222142
  *
  * @package OCA\Social\Migration
  */
-class Version0001Date20181208185242 extends SimpleMigrationStep {
+class Version0001Date20181211222142 extends SimpleMigrationStep {
 
 
 	/**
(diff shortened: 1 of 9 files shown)


lib/Migration/Version0001Date20181211222142.php
```diff
@@ -62,27 +66,28 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		$table = $schema->getTable(CoreRequestBuilder::TABLE_CACHE_ACTORS);
 		$table->changeColumn(
 			'source', [
-						'notnull' => true,
-						'length'  => CoreRequestBuilder::SOURCE_LENGTH,
+						'notnull' => false,
+						'type'    => Type::getType(Type::TEXT)
```
nickvergessen: if I remember correctly changing type is not supported on all DBMS.
So you have to create a new column and copy over the values and then delete the old one.

See https://github.com/nextcloud/server/pull/12275/files#diff-1307110df255230fc590b38bbc452bc1 for the simplest example

daita: improved

daita:
multiple changes regarding the modification of the type of some field:

List of the fields: https://github.com/nextcloud/social/blob/5e0cb1f92d00b8a605cb08cd92946cde7e0314b0/lib/Migration/Version0001Date20181219000001.php#L58-L75:

- most of the already big varchar are now text,
- some varchar (length > 1000) are now varchar(4000)

The process:

- step 1: creating a new _field_copy_ and copy the content.
- step2: removing old _field_
- step3: creating a new _field_ and copy the ... copy.
- step4: removing old _field_copy_

PR #226, opened 2018-12-13
[huge] rework on the AP Objects

daita:
Rework were supposed to come with 0.2.0, but I am fine for 0.1.1.

- Parsing is split between processIncomingRequest and processResult.
- AP Objects are now split in 'Activity', 'Actor', 'Object'
- +activity()

The idea is that when processing an incoming request, 'Activity' Object will take care of it and execute the 'activity()' method on child object. 
All operations are managed by the child, depending of the type of the Activity.  
Also, some object can be hybrid and now be used as an Activity or an Object (ie. Follow)  
Checking Origin and validating structure of the JSON will be slightly easier this way.


jancborchardt:
cc @nextcloud/social if you already want to check this out :)

lib/AP.php
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+
+	/**
+	 *
+	 */
+	public static function init() {
+		$ap = new AP();
+		try {
+			$ap->acceptInterface = \OC::$server->query(AcceptInterface::class);
```
juliushaertl: Can't we use proper dependency injection here?

lib/Controller/AccountController.php
```diff
@@ -121,6 +121,9 @@ public function create(string $username): DataResponse {
 
 
 	/**
+	 *
+	 * // TODO - is it still used ? maybe using info from LocalController !?
```
juliushaertl: :+1: Yes, lets migrate this to the local controller in a follow up PR.

lib/Interfaces/Object/DocumentInterface.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @param MiscService $miscService
+	 */
+	public function __construct(
+		CacheDocumentsRequest $cacheDocumentsRequest, MiscService $miscService
+	) {
+		$this->cacheDocumentsRequest = $cacheDocumentsRequest;
+		$this->miscService = $miscService;
+	}
+
+//
```
juliushaertl: Let's remove those comments if we don't need the cod anymore :wink:

lib/Interfaces/Activity/DeleteInterface.php
```diff
@@ -0,0 +1,148 @@
...
+		$object = $item->getObject();
+
+		try {
+			$service = AP::$activityPub->getInterfaceForItem($object);
+			$service->delete($object);
+		} catch (UnknownItemException $e) {
+		}
+	}
+
+//
```
juliushaertl: This is probably not needed as well.

lib/Interfaces/Activity/FollowInterface.php
```diff
@@ -0,0 +1,222 @@
...
+	/**
+	 * @param Follow $follow
+	 */
+	public function confirmFollowRequest(Follow $follow) {
+		try {
+			$remoteActor = $this->cacheActorService->getFromId($follow->getActorId());
+
+			$accept = new Accept();
+			// TODO: improve the generation of the Id
+			$accept->setId($follow->getObjectId() . '#accepts/follows/' . rand(1000, 100000000));
```
juliushaertl: Please use generate() from ISecureRandom

lib/Interfaces/Actor/PersonInterface.php
```diff
@@ -0,0 +1,142 @@
...
+
+
+	/**
+	 * @param ACore $activity
+	 * @param ACore $item
+	 */
+	public function activity(Acore $activity, ACore $item) {
+		/** @var Person $item */
+		if ($activity->getType() === Update::TYPE) {
+//			$this->miscService->log('### UPDATE PERSON !' . json_encode($item));
```
juliushaertl: Debugging leftover?

lib/Interfaces/Object/DocumentInterface.php
```diff
@@ -0,0 +1,244 @@
...
+		/** @var Document $item */
+		$this->cacheDocumentsRequest->save($item);
+	}
+
+
+	/**
+	 * @param ACore $item
+	 */
+	public function delete(ACore $item) {
+//		$this->cacheDocumentsRequest->delete($item);
```
juliushaertl: I think it should be called here, right?

lib/Model/ActivityPub/Activity/Add.php
```diff
@@ -28,37 +28,54 @@
...
 
 
 	/**
-	 * @param ACore $item
+	 * @return array
 	 */
-	public function delete(ACore $item);
-
+	public function jsonSerialize(): array {
+		return array_merge(
```
juliushaertl: No need for a method if it is just returning the parents array.

lib/Model/ActivityPub/Activity/Block.php
```diff
@@ -0,0 +1,81 @@
...
+	public function import(array $data) {
+		parent::import($data);
+	}
+
+
+	/**
+	 * @return array
+	 */
+	public function jsonSerialize(): array {
+		return array_merge(
```
juliushaertl: No need for a method if it is just returning the parents array.

lib/Model/ActivityPub/Activity/Remove.php
```diff
@@ -0,0 +1,81 @@
...
+	public function import(array $data) {
+		parent::import($data);
+	}
+
+
+	/**
+	 * @return array
+	 */
+	public function jsonSerialize(): array {
+		return array_merge(
```
juliushaertl: No need for a method if it is just returning the parents array.

lib/Model/ActivityPub/Activity/Update.php
```diff
@@ -0,0 +1,81 @@
...
+	public function import(array $data) {
+		parent::import($data);
+	}
+
+
+	/**
+	 * @return array
+	 */
+	public function jsonSerialize(): array {
+		return array_merge(
```
juliushaertl: No need for a method if it is just returning the parents array.

lib/Service/ActivityService.php
```diff
@@ -212,14 +189,14 @@ public function getItem(string $id): ACore {
 		}
 
 		$requests = [
-			$this->notesRequest
+			'Note'
```
juliushaertl: Any reason this is an array that is iterated over and not just one call? Also lets use Note::TYPE instead of the hardcoded string.

lib/Service/ActivityService.php
```diff
@@ -471,144 +411,157 @@ private function getAuthorFromItem(Acore $activity): string {
...
-			if ($key === '(request-target)') {
-				continue;
-			}
-
-			$estimated .= "\n" . $key . ': ' . $request->getHeader($key);
-		}
-
-		return $estimated;
-	}
+//	/**
```
juliushaertl: Can we remove this?

lib/Service/ActorService.php
```diff
@@ -87,275 +64,243 @@ class ActorService {
...
 
-	/**
-	 * @param $username
-	 */
-	private function checkActorUsername($username) {
-		$accepted = 'qwertyuiopasdfghjklzxcvbnm';
-
-		return;
-	}
+//	/**
```
juliushaertl: :fire: Let's remove that as well

lib/Service/CurlService.php
```diff
@@ -68,6 +76,83 @@ public function __construct(ConfigService $configService, MiscService $miscServi
...
+	 * @throws RequestException
+	 * @throws SocialAppConfigException
+	 * @throws RedundancyLimitException
+	 * @throws UnknownItemException
+	 * @throws InvalidOriginException
+	 */
+	public function retrieveAccount(string $account): Person {
+		$account = $this->withoutBeginAt($account);
+
+		if (strstr(substr($account, 0, -3), '@') === false) {
```
juliushaertl: This check doesn't seem right `@test` will also be a valid account then. Having the additional null check after the explode would be fine I think.

lib/Service/FollowService.php
```diff
@@ -27,45 +27,57 @@
...
-use OCA\Social\Service\ConfigService;
-use OCA\Social\Service\MiscService;
 
+class FollowService {
+
+
+	use TArrayTools;
+
+
+	const REQUEST_INBOX = 1;
```
juliushaertl: Can we unify this with the constants from ActivityService? We have the same ones over there.

lib/Service/FollowService.php
```diff
@@ -78,27 +90,29 @@ class FollowService implements ICoreService {
...
 		$this->followsRequest = $followsRequest;
-		$this->personService = $personService;
 		$this->activityService = $activityService;
+		$this->cacheActorService = $cacheActorService;
 		$this->configService = $configService;
 		$this->miscService = $miscService;
 	}
 
 
+	//
```
juliushaertl: :fire:

lib/Service/ImportService.php
```diff
@@ -109,94 +71,54 @@ public function __construct(
...
-		try {
-			/** @var Document $icon */
-			$icon = $this->importFromData($this->getArray('icon', $data, []), $item);
-			$item->setIcon($icon);
-		} catch (UnknownItemException $e) {
-		}
-
-		return $item;
-	}
+//
```
juliushaertl: Still needed?

lib/Service/SignatureService.php
```diff
@@ -0,0 +1,266 @@
...
+use OCA\Social\Model\ActivityPub\Actor\Person;
+use OCP\IRequest;
+
+class SignatureService {
+
+
+	use TArrayTools;
+
+
+	const DATE_FORMAT = 'D, d M Y H:i:s T';
```
juliushaertl: We should have this format defined in a wider scope, this is also duplicate around some places.

lib/AP.php
daita: This is the only solution I found to avoid cross-injecting of the Parser (including all Interface, and their service on complex activitypub objects. The parser needs to be a static object, and it makes all the parsing and the creation of an ActivityPub object a lot easier.

lib/Interfaces/Activity/FollowInterface.php
daita: using the generateUniqueId() of the model that generate a uuid

lib/Service/SignatureService.php
daita: yup, fixed !

lib/Service/FollowService.php
daita: fixed

PR #240, opened 2018-12-18
small example of a widget

daita:


jancborchardt:
Nice! :) @daita what should we display in the widget? I guess:
- the composer
- the last ~3 posts of Home, or whatever fits in the size people use

jancborchardt:
Outdated since we have the new Dashboard approach – let’s continue in the issue for that instead: https://github.com/nextcloud/social/issues/983

Issue #229, opened 2018-12-14
Browsing to account results in "File not found" error page


mrstegeman:
**Describe the bug**

After creating the post, I click on the link above it in the timeline and end up with a "File not found" error, instead of the account page.

**To Reproduce**

1. Go to Social app
2. Create a public post
3. Go to any of the timeline pages, i.e. Profile, Local timeline, or Global timeline
4. Click on link above post, i.e. https://myhost.com/index.php/apps/social/@myuser

**Expected behavior**

Account/profile page should be displayed

**Screenshots**

![screenshot_2018-12-14 nextcloud](https://user-images.githubusercontent.com/457381/50019478-ba5e0280-ffca-11e8-91e4-ad353f707c20.png)

**Client details:**

 - OS: macOS
 - Browser: Firefox
 - Version: 64
 - Device: Desktop

<details>
<summary>Server details</summary>

**Social app version:** 0.1.0

**Operating system:** Official Docker image

**Web server:** Apache

**Database:** mariadb 10.3.11

**PHP version:** 7.2

**Nextcloud version:** 15.0.0

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
{"reqId":"zN5FZ7w1xqxfw63gxhIY","level":0,"time":"2018-12-14T18:06:41+00:00","remoteAddr":"172.20.0.2","user":"michael","app":"no app in context","method":"GET","url":"\/index.php\/apps\/social\/@michael","message":"No cache entry found for \/appdata_ocvfkofw8r5z\/theming\/images\/background (storage: local::\/var\/www\/html\/data\/, internalPath: appdata_ocvfkofw8r5z\/theming\/images\/background)","userAgent":"Mozilla\/5.0 (Macintosh; Intel Mac OS X 10.13; rv:60.0) Gecko\/20100101 Firefox\/60.0","version":"15.0.0.10"}
[...]

Issue #172, opened 2018-12-08
Issues with public profile page


jancborchardt:
- [x] The "public" page of a user only shows when you are logged in. Which defeats the purpose of course. 
    ![social public page logged in](https://user-images.githubusercontent.com/925062/49678412-51cfcc80-fa84-11e8-8f69-f6e077df2379.png)
    In private mode or when not logged in, it’s empty with some errors:
    ![social public page private mode](https://user-images.githubusercontent.com/925062/49678411-51cfcc80-fa84-11e8-86c9-598335dc6ad6.png)
- [x] In the header it says "Social `username`". It should just say "Social" as the profile is directly below.
- [x] In the title it only says "Nextcloud". It should say "`Name` - Social - Nextcloud" as we do in other apps too. The most specific thing on the left.
- [x] The tab "Posts" should shown and selected by default.

schiessle:
btw, wouldn't it be nice to use `https://<your-nextcloud>/@user` as profile page, without the `apps/social` part, at least as a redirect? This would make it easier to discover and remember user profiles and most people know this pattern already from Mastodon.

jancborchardt:
@schiessle great point – could you open a separate issue for it? :)

mknet:
@jancborchardt Is there anything I have to configure to activate public profiles? I always get the login form, though I'm using https://domain/index.php/apps/social/@user/

Thanks for your help

daita:
please create a new issue for this one !

mknet:
Okay, will do so.

octavsly:
> Okay, will do so.

what is the new issue number?

Issue #219, opened 2018-12-12
update cache_actors Update/Person 


daita:
The app should manage the entering request when some edit his profile:

```
{  
  "type": "Update",  
  "object": {
    "type": "Person",
  }
}
```

and update the cache.

PR #244, opened 2018-12-19
Public page frontend fixes

juliushaertl:
* Resolves: #172 #229
* Target version: master 

### Summary
- [x] The "public" page of a user only shows when you are logged in. Which defeats the purpose of course. 
- [x] In the header it says "Social `username`". It should just say "Social" as the profile is directly below.
- [x] In the title it only says "Nextcloud". It should say "`Name` - Social - Nextcloud" as we do in other apps too. The most specific thing on the left.
- [x] The tab "Posts" should shown and selected by default.
image
- [x] Added proper error page for non-existing users
image


lib/Controller/LocalController.php
```diff
@@ -628,6 +632,15 @@ public function documentsCache(array $documents): DataResponse {
 	 * @throws AccountDoesNotExistException
 	 */
 	private function initViewer(bool $exception = false) {
+		// only initialize viewer data for logged in users
+		if ($this->userId === null) {
```
juliushaertl: @daita Maybe you can help here. Is there a good way to still get details for actors if no viewerId is set?

lib/Controller/LocalController.php
```diff
@@ -239,6 +239,7 @@ public function streamNotifications($since = 0, int $limit = 5): DataResponse {
 	/**
 	 * @NoAdminRequired
 	 * @NoSubAdminRequired
+	 * @PublicPage
```
juliushaertl: This allows guest users to fetch the account stream.

daita: not sure we want this. We already have a streamAccount ?

lib/Controller/LocalController.php
daita: what about using LocalController::accountInfo() and add false as a parameter of `initViewer`

```
			$this->initViewer(false);
```

lib/Controller/LocalController.php
juliushaertl: This is actually just adding the PublicPage annotation to the streamAccount method (GitHub diff is a bit strange :wink: )

lib/Controller/LocalController.php
juliushaertl: Hm let me check that again, since we actually don't need the following/follower (if the current user is following) details, just the number of post/followers/following.

lib/Controller/AccountController.php
```diff
@@ -1,173 +0,0 @@
...
-	 *
-	 * @NoAdminRequired
-	 * @NoSubAdminRequired
-	 *
-	 * @param string $username
-	 *
-	 * @return DataResponse
-	 */
-	public function create(string $username): DataResponse {
-		try {
```
daita: I would keep this method somewhere: we will have a request someday to let users creating their account manually, and choose their username.

jancborchardt:
@mrstegeman could you review this as it fixes the issue https://github.com/nextcloud/social/issues/229 you reported? :slightly_smiling_face:  Thanks a lot!

lib/Controller/AccountController.php
juliushaertl: We still have it in the git history. Also it is not that much code, so I'm happy to restore it, if we need it again :wink:

lib/Controller/LocalController.php
```diff
@@ -250,6 +251,7 @@ public function streamAccount(string $username, $since = 0, int $limit = 5): Dat
 			$this->initViewer();
 
 			$account = $this->cacheActorService->getFromLocalAccount($username);
+			$account->setCompleteDetails(true);
```
daita: not sure it is useful, what are you trying to achieve ?

jancborchardt:
@daita @juliushaertl could we release 0.1.1 soon btw? We keep adding more stuff into it, what of https://github.com/nextcloud/social/milestone/5 is really needed, and what should we just move to 0.1.2?

daita:
_Release is coming_ tomorrow or this week-end

Let me finish:
- [ ] #217 
- [ ] #219 
- an Exception from Sentry

jancborchardt:
@daita alright! Moved all open ones except #217 and #219 over to 0.1.2 🥚👩‍🏭

PR #249, opened 2018-12-20
Cleaning: unused import and composer lib

daita:


composer.json
```diff
@@ -12,8 +12,5 @@
   "require": {
     "daita/my-small-php-tools": "dev-master",
     "digitalbazaar/json-ld": "0.4.7"
-  },
-  "require-dev": {
-    "jakub-onderka/php-parallel-lint": "^1.0"
```
juliushaertl: This is actually used for the php syntax check in the CI :wink:

daita: so that where it's coming from !

I have some conflict with JsonSerialize with it, is it normal ?

juliushaertl: What kind of conflict?

Issue #238, opened 2018-12-17
Add "rsa-sha256" as "algorithm" to the HTTP signature


annando:
Currently the HTTP signature is missing the "algorithm" field.

Please add ```algorithm="rsa-sha256"``` to the HTTP signature fields, like most other implementations are doing.

jancborchardt:
@annando would you be interested in contributing this fix? :slightly_smiling_face:  The development setup instructions are at https://github.com/nextcloud/social#development-setup, let us know if you need any help!

annando:
Currently I'm still running with version 14, since the update process hadn't offered me the update yet. And I'm running it on my NAS, I'm not sure if I can setup a development environment there.

jancborchardt:
Fixed through https://github.com/nextcloud/social/pull/251

PR #251, opened 2018-12-20
signed header generation moved to SignatureService

daita:
Moving the generation of the header to sign a request from ActivityService to SignatureService

### DIFF
--- a/lib/Service/SignatureService.php
+++ b/lib/Service/SignatureService.php
@@ -31,9 +31,12 @@ namespace OCA\Social\Service;
 
 
 use daita\MySmallPhpTools\Exceptions\MalformedArrayException;
+use daita\MySmallPhpTools\Model\Request;
 use daita\MySmallPhpTools\Traits\TArrayTools;
 use DateTime;
 use Exception;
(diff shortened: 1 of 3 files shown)


daita:
Also, fixing #238 by filling the header, and checking algorithm on incoming request

jancborchardt:
cc @annando for review as this fixes the issue https://github.com/nextcloud/social/issues/238 you reported :)

lib/Service/SignatureService.php
```diff
@@ -225,9 +269,15 @@ private function checkRequestSignature(IRequest $request): string {
 		$signed = base64_decode($sign['signature']);
 		$estimated = $this->generateEstimatedSignature($headers, $request);
 
+		$algorithm = 'sha256';
+		if ($this->get('algorithm', $sign, '') === 'rsa-sha256') {
```
annando: Possibly you could check for ```'rsa-sha512'``` here as well, like we do it here:
https://github.com/friendica/friendica/blob/2018.12-rc/src/Util/HTTPSignature.php#L421-L427

daita: should work now

PR #252, opened 2018-12-20
./occ social:reset

### LINKED ISSUE #233: Change / reset / server configure federated cloud ID
**Describe the bug**
I have used both local IP 192.168.xx.yy as well as remote IP my.domain.com to log onto my nextcloud.
Sharing under settings reflects the way I have logged on ie john@192.168.xx.yy if logged on via  local ip or john@my.domain.com if logged on via the web.

social seems stuck on the local ip ie john@192.168.xx.yy and I do not seem to be able to change it - this is independent of log on method.

**To Reproduce**
I cannot change the settings or cannot find a way to force the format of the federated ID anywhere so I cannot do a reproduce

**Expected behavior**
User should be allowed to set (& fix) a federated cloud ID in either social or system settings
and/or
Admin should be able to fix permissible federated cloud IDs, for instance limit domains

- - I am not skilled enough to find how I can fix this by "forcing the database" I have had a little dig  through the mysql tables but cannot find if it is stored as a fixed setting there, 

Using NC 15.0 on apache2 ubuntu 18.04

daita:
Add a command for admin to reset and change their cloud base address

Fixes #233

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -47,6 +47,7 @@
 	<commands>
 		<command>OCA\Social\Command\CacheRefresh</command>
 		<command>OCA\Social\Command\NoteCreate</command>
+		<command>OCA\Social\Command\Reset</command>
 		<command>OCA\Social\Command\QueueStatus</command>
 		<command>OCA\Social\Command\QueueProcess</command>
 	</commands>
--- a/lib/Command/Reset.php
+++ b/lib/Command/Reset.php
@@ -1,0 +1,147 @@
+<?php
+declare(strict_types=1);
+
+
+/**
+ * Nextcloud - Social Support
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
+ * @author Maxence Lange <maxence@artificial-owl.com>
+ * @copyright 2018, Maxence Lange <maxence@artificial-owl.com>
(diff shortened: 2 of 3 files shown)


daita:
selection_052

jancborchardt:
@john-2000 @Lars1963 @mcnesium please review as this fixes the issue you were involved in at https://github.com/nextcloud/social/issues/233 :)

daita:
As shown on the screenshot, at the end of the reset, you have the possibility to change the url. 

Also, changing the addiress while using the social app will not be harmless for the whole decentralized part of ActivityPub. This is why, at this early point of the developement of the app, a reset is considered as mandatory when changing the url.

Lars1963:
Works for me.

john-2000:
Hi 
the reset worked, in order to then make posts I had to redownload the official app version, without posts but it is well possible that I mucked something up while patching.

jancborchardt:
@daita could you also document this in the readme?

violoncelloCH:
@jancborchardt enhancement in #266

mracreep:
@noplanman how did u get the social app reinstalled after that?
For me the tables are not recreated... :/
I deleted the following:
* Appfolder of social
* tables oc_social*
* social entries in oc_appconfig 
* social entreis in oc_migrations...

PR #255, opened 2018-12-21
htmlentities encode/decode

daita:
Encode on post creation,
Decode on object parsing.


daita:
Yup yup, you're right, it works only remotely, not locally :]

daita:
@juliushaertl this new PR should not validate content of a new post during the import/creation, but when it's extracted from database.

The rest should still be checked during import.

PR #266, opened 2018-12-21
document "occ social:reset" in readme

violoncelloCH:
* Resolves: https://github.com/nextcloud/social/pull/252#issuecomment-449133116
* Target version: master 

### Summary
short documentation on the "occ social:reset" for the readme


jancborchardt:
Thanks a lot, good stuff! :) We always need to document stuff like this.

PR #274, opened 2018-12-24
Using signature date as 'creation' date

daita:
Feature #219

We keep the current time when requesting for data of a Perso.
On incoming request, we assign the date used in the signature (request and/or object signing)
On update from remote instance (incoming request Update/Person), we use this signature as date and compare if current Person in cache is older than the request.

### DIFF
--- a/lib/Service/SignatureService.php
+++ b/lib/Service/SignatureService.php
@@ -161,6 +161,8 @@ class SignatureService {
 	/**
 	 * @param IRequest $request
 	 *
+	 * @param int $time
+	 *
 	 * @return string
 	 * @throws InvalidOriginException
 	 * @throws InvalidResourceException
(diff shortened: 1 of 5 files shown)


lib/Model/ActivityPub/Actor/Person.php
```diff
@@ -475,7 +475,14 @@ public function import(array $data) {
 	 */
 	public function importFromDatabase(array $data) {
 		parent::importFromDatabase($data);
-		$this->setPreferredUsername($this->validate(self::AS_USERNAME, 'preferred_username', $data, ''))
+
+		$dTime = new DateTime($this->get('creation', $data, 'yesterday'));
+		//$dTime->format(SignatureService::DATE_FORMAT);
```
juliushaertl: This is not needed.

lib/Service/SignatureService.php
```diff
@@ -209,8 +212,13 @@ public function checkObject(ACore $object): bool {
 			$signature->import(json_decode($object->getSource(), true));
 			$signature->setPublicKey($this->retrieveKey($actorId));
 			if ($signature->verify()) {
+
+				$dTime = new DateTime($signature->getCreated());
+				$dTime->format(self::DATE_FORMAT);
```
juliushaertl: No need for the format call.

Issue #269, opened 2018-12-23
SignatureException behind Apache proxy


riemy:
**Describe the bug**
Everytime an external service tries to communicate with Social the messages cannot be processed and result in a **SignatureException** as stated by a log entry like seen below.

**My setup**
Nextcloud running in a docker container <-- http --> Apache with proxy setup 443 to 80 and an official subdomain <-- https --> Internet

**First approach of a solution**
So I debugged a little and found, that the signature comparison done in `SignatureService->generateEstimatedSignature()` evaluated the header variable 'Host', which was filled with the name and port of my docker container. 

Thinking, that this might be wrong, I took a peek at the available request headers at this point of execution. Apache put the "real", external hostname into `X-Forwarded-Host`, so I tried building the estimated Signature with this variable instead. This worked, although I can't say that I fully understand the implications of my hack. Here is what my method looks like now (added the second if-block in the loop):

```php
	private function generateEstimatedSignature(string $headers, IRequest $request): string {
		$keys = explode(' ', $headers);
		$target = '';
		try {
			$target = strtolower($request->getMethod()) . " " . $request->getRequestUri();
		} catch (Exception $e) {
		}
		$estimated = "(request-target): " . $target;
		foreach ($keys as $key) {
			if ($key === '(request-target)') {
				continue;
			}
			if ($key === 'host' && isset($request->getHeader("X-Forwarded-Host")))
			{
				$estimated .= "\n" . $key . ': ' . $request->getHeader("X-Forwarded-Host");
				continue;
			}
			$estimated .= "\n" . $key . ': ' . $request->getHeader($key);
		}
		return $estimated;
	}
```

I can't even say, if this is maybe just a configuration problem of my setup, or if it is really a bug. Or my solution might lead to some horrible security issues. Or it works only on my special setup. All I know is that it worked after that change. What do you think?

**To Reproduce**
1. Send a direct Mastodon message to my nextcloud account.
2. Results in SignatureException like below.

**Expected behavior**
The message is being processed, received and displayed.

<details>
<summary>Server details</summary>
**Social app version:** 0.1.1

**Operating system**: Gentoo Linux

**Web server:** Apache 2.4

**Database:** MySQL

**PHP version:** 7.2.13

**Nextcloud version:** 15.0.0.10

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
[...]

PR #279, opened 2018-12-26
switch to getInsecureServerHost

daita:
This would help when using Apache Proxy.


daita:
I see no real downside to assume that the host IS unique, and generate the estimated signature based on this host.

riemy:
I can also confirm this working for my setup (also #269).

PR #282, opened 2018-12-29
managing Notifications in the notification steam

daita:
A new object will be stored in the server_notes database. This object is returned in the Notification Stream (and only there right now). The Type is 'SocialAppNotification' and the text is in the summary key. actor_info is filled also with the actor at the origin of the notification.

Currently, only reading is available, the creation/saving of the object in database is not done yet.


daita:
@juliushaertl : most of the back-end is done.

SocialAppNotification item is generated in the table _server_notes_

```
{
  "@context": [
    "https://www.w3.org/ns/activitystreams"
  ],
  "id": "https://mastodon.xyz/056a9eb3-1b8d-43dd-8111-cc11fcee399f/notification",
  "type": "SocialAppNotification",
  "attributedTo": "https://mastodon.xyz/users/testing",
  "to": "https://test.artificial-owl.com/apps/social/@cult",
  "summary": "{actor} is following you",
  "local": true
}
```

It is returned in the notification stream like a _Note_ item, including actor_info:

selection_053


For 0.2, we should rename the table _**server_notes**_ to _**server_streams**_ as it includes all type of object we displays in the streams.

daita:
some parts should be edited to fit the Model/Stream from #374

PR #283, opened 2018-12-29
[huge] rewrite of the exceptions while requesting remote instance

daita:
The idea is to split all possible errors when requesting a remote instance in 4 groups to improve/manage the queue:

- Network issues - Request stays in the queue and a background process will retry later.
- Server issues - Request stays in the queue and a background process will retry later
- Content issues - Request is deleted.
- ActivityPub issues - Request is deleted.

Also, the caching of documents is now using the CurlService::request() to download content

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -12,12 +12,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/daita/my-small-php-tools.git",
-                "reference": "29754f18951856a22c0fd5fc388b6162ea98fe8a"
+                "reference": "0baac1f399b257b00fdc2eafb754396ffa1892f4"
             },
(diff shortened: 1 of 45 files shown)


daita:
Following this PR, a new task should be created to implement a better followup on exception if the request is initiated by an incoming entry. 
Right now if the remote instance is not available, the incoming request is ignored.
Like getting the public key on a bounced message from an author not in cache already:

- add a 'verified' field in the right table in the database.
- save the data from the incoming request but check the 'verified' field only if the public key is available and the object is correctly signed
- add a background task that keep trying to get the public key and check the signed object when public key is available.
- update the 'verified' field, and check the field is verified when retrieving any stream.

PR #291, opened 2019-01-02
Support blind keys rotation

daita:
- while verifying remote object, on fail, get new signature.
- local account got their keys refreshed every 7 days


juliushaertl:
@daita Mind to rebase?

daita:
I'll see if I can do that without breaking everything, give me an hour

PR #295, opened 2019-01-02
Hashtags

daita:
Hashtags in post will be saved in server_notes/tags as a JSON, search on a specific tag will be done within this field.

A new table (server_tags) will be updated by a background job and will store all hashtags used in post, and data about its trend/use within the last minutes/hours/days/...


daita:
@juliushaertl we will need to 

- display result of a search for tags (like the search for accounts)
- a page that display posts relative to a specific hashtag: https://cloud.example.net/apps/social/timelines/tag/myhashtag

daita:
The stream should returns post related to an hashtag: 

selection_055

daita:
There is a new endpoint for the search: both search (during Post creation, and searchbar) should use it.

The endpoint should returns everything related to Accounts/Hashtags (and content in the future):

- based on first char: if '@', will returns only accounts. if '#', will returns only hashtags, else will returns hashtags and accounts.
- there is no pagination/limit right now. This needs to be implemented.
selection_055
selection_057

daita:
@juliushaertl  the back-end part is done, we just need some Vue magic now !

juliushaertl:
Timeline for hashtags:
image

Search:
image

I'd say this is ready for review as a first iteration.

jancborchardt:
Very nice with the activity visualization @juliushaertl! :) (@daita graphs don’t need to be sexualized though. ;)

daita:
I will have a look, also fixing the conflicts, just that I am in the middle of a huge PR :-)

Remember that the search is using some cache, so you might need to refresh the cache with `./occ social:cache:refresh`

https://github.com/nextcloud/social/pull/295/files#diff-7ac0a5485292d7aa02d0cad31b4b0418R127

violoncelloCH:
> Remember that the search is using some cache, so you might need to refresh the cache with `./occ social:cache:refresh`

ahh, thank you for the hint... now search is working and finding all tags!

daita:
conflict should be solved by rebase

juliushaertl:
@daita Rebased and added a limit of 25 results to the search for now. I'm not sure if we actually need to paginate this anyway, since if there is no proper result in the first 25 matches, the user should just be more specific in the search query.

daita:
@juliushaertl yes, nice !

juliushaertl:
@daita @violoncelloCH Please review, so we can get this merged :wink:

daita:
You know that I cannot review my own PR :D

PR #304, opened 2019-01-04
Compat friendica

daita:


lib/Model/LinkedDataSignature.php
```diff
@@ -237,19 +260,23 @@ public function verify(): bool {
...
+		$hashObject = $this->hashedCanonicalize($this->getObject());
 
 		$algo = OPENSSL_ALGO_SHA256;
 		if ($this->getType() === 'RsaSignature2017') {
 			$algo = OPENSSL_ALGO_SHA256;
 		}
 
-		if (openssl_verify($hash, $signed, $this->getPublicKey(), $algo) === 1) {
+		if (openssl_verify(
+				$hashHeader . $hashObject, base64_decode($this->getSignatureValue()),
```
juliushaertl: Maybe we should catch a failure in base64_decode as well

daita: done

PR #323, opened 2019-01-10
Add UI to reply to a post and delete

### LINKED ISSUE #155: Delete your own post
We need to add a way to delete  your own post on the UI, and manage this on the back-end

### LINKED ISSUE #158: [UI] Ability to reply to a Post.
Theoretically, the back-end is already able to do that, an implementation on the UI should be enough. The idea is to fill the 'replyTo' field with the Id of the original post during the creation of the Post

juliushaertl:
Fixes #158 
Fixes #155

image

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -1,0 +1,1 @@
+@include icon-black-white('reply', 'social', 1);
--- a/img/reply.svg
+++ b/img/reply.svg
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.w3.org/2000/svg" height="16" width="16" version="1.1" xmlns:cc="http://creativecommons.org/ns#" xmlns:dc="http://purl.org/dc/elements/1.1/">
+ <g transform="matrix(-1,0,0,1,16,-1036.4)">
+  <path d="m1 1051.4s0.39147-7.8025 7-10v-4l7 7-7 7v-4c-5.1157 0-7 4-7 4z"/>
+ </g>
+</svg>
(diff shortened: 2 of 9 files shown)


jancborchardt:
Replying should be outside of the 3-dot menu (much like Favorite and Repost in the future) as it’s a core action of the app. Should be an icon with text below the post text, like on Mastodon etc.

As icon, the reply arrow is good (which you use in the embed too)  and as text "Reply" is enough. (Not the comment icon and "Reply to post" as in the menu. :)

jancborchardt:
Also – wow, very nice! :balloon: Wasn’t expecting this already now, great work @juliushaertl!

juliushaertl:
@jancborchardt 
image

daita:
Thanks for this report, i'll have a look on a fresh install

daita:
> looks nice!
> however:
> 
>     * the three dot menu is also shown on posts from other people where deletion is not possible and it's empty therefore... we shouldn't show the three dots if there is no option in it:
>       image
> 

Confirming this issue, empty box under the arrow. @juliushaertl 

>     * also I get some js console errors, but that's probably an issue on my side with the backend; occ social:cache:refresh throws an unhandled exeption `Error: Call to a member function blindKeyRotation() on null in /home/jonas/nextcloud/apps/social/lib/Command/CacheRefresh.php:112`
> 
>     * I also can't see the reply posts anywhere

There is some issue with the actor_info, you might see something about it in your JS console: #404 
The issue with the CacheRefresh is not from this PR. I am working on this.

jancborchardt:
Looks nice @juliushaertl – only the menu is now left-aligned, should be centered below the 3-dot icon like before. :)

violoncelloCH:
I'm also wondering if there should be any kind of indication in the timeline that one post is a reply to another... or will this be done in a following PR?
for me replies look like normal posts and are added at the top of the timeline

daita:
no; but you're right that this should be implemented (in another PRs)

juliushaertl:
@jancborchardt Adressed your comments, this is what it looks like now:

image

daita:
@juliushaertl  : if you don't want to quickly fix the last issue reported by @violoncelloCH, we can we merge and create a bug report for later !

daita:
merging, so it will be easier to work on the boost option #374

PR #340, opened 2019-01-16
Refresh some Actor details in cache on Follow events

daita:
and when a new post is created.


tagmention.html
```diff
@@ -0,0 +1,6 @@
+<p><span class="h-card"><a href="https://mastodon.xyz/@nextcloud"
```
juliushaertl: Unrelated

PR #341, opened 2019-01-16
instructions about Emojis

daita:


jancborchardt:
Huh, is this something that needs to be done manually on every installation? Cause Emojis should ideally just work – is there no way for us to do it automatically?

And how does the Talk app do it @nickvergessen?

nickvergessen:
@jancborchardt we just don't talk about it. It's default for new installations when the mysql/maria version supports it. all other dbs support it out of the box.

But well, if you'd take a look at https://docs.nextcloud.com/server/stable/admin_manual/configuration_database/mysql_4byte_support.html you would see that we can no do this automatically. Because the webserver can luckily never write /etc/mysql/mysql.ini and restart your database server...

9662:
> Because the webserver can luckily never write /etc/mysql/mysql.ini and restart your database server.

This is completely untested, but I suspect you could check the value of `innodb_file_per_table` (via `select @@innodb_file_per_table;`) and warn the user in the admin page along with the other setup/config warnings?

Issue #306, opened 2019-01-04
Failed to load more timeline entries


emersonkfuri:
**Describe the bug**
'Failed to load more timeline entries' caused by "OCA\\\\Social\\\\Exceptions\\\\CacheActorDoesNotExistException". CacheActorsRequest->getFromLocalAccount receives @user@cloud.example.com notation.

**To Reproduce**
1. Go to 'Direct messages'
2. Click on '@user@cloud.example.com'
3. See error 'Failed to load more timeline entries'

**Social app version:**
0.1.2

**Nextcloud version:**
15.0.0

#### Nextcloud log (data/nextcloud.log)
```
==> /opt/nextcloud/data/nextcloud.log <==
{"reqId":"5AUpqoX3slpdQV4v2g2P","level":2,"time":"2019-01-04T15:01:25+00:00","remoteAddr":"XXX.XXX.XXX.XXX","user":"emerson","app":"no app in context","method":"GET","url":"\/apps\/social\/api\/v1\/account\/emerson@conecta.ufmg.br\/stream?limit=25&since=1546614085","message":"500 - {\"status\":-1,\"exception\":\"OCA\\\\Social\\\\Exceptions\\\\CacheActorDoesNotExistException\",\"message\":\"\"}","userAgent":"Mozilla\/5.0 (X11; Fedora; Linux x86_64) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/73.0.3642.0 Safari\/537.36","version":"15.0.0.10"}
```

### My local resolution
```
--- a/CacheActorsRequest.php	2019-01-04 13:07:33.710196095 -0200
+++ b/CacheActorsRequest.php	2019-01-04 13:08:05.061402029 -0200
@@ -220,6 +220,9 @@
 	 */
 	public function getFromLocalAccount(string $account): Person {
 		$qb = $this->getCacheActorsSelectSql();
+		if (strrpos($account, '@')) {
+			$account = substr($account, 0, strrpos($account, '@'));
+		}
 		$this->limitToPreferredUsername($qb, $account); 
 		$this->limitToLocal($qb, true);
 		$this->leftJoinCacheDocuments($qb, 'icon_id');
```

daita:
interesting, can you create a PR for this fix ?

PR #344, opened 2019-01-16
Failed to load more timeline entries

emersonkfuri:
* Resolves: #306 
* Target version: master


emersonkfuri:
> Thanks for your PR!
> 
> However let's keep the CacheActorsRequest clean. Could you implement your code within
> 
> [social/lib/Service/CacheActorService.php](https://github.com/nextcloud/social/blob/2f58b914081bc5d7beee419b974cec6029948a36/lib/Service/CacheActorService.php#L168)
> 
> Line 168 in [2f58b91](/nextcloud/social/commit/2f58b914081bc5d7beee419b974cec6029948a36)
> 
>  	public function getFromLocalAccount(string $account): Person { 
> instead ?

Done! Sorry by the mess... Not used to work over here... :D

PR #352, opened 2019-01-19
Incoming Attachments.

daita:
- document attached to remote Notes are cached,
- attachments list is rewrited to fit the cache,
- attachments are attached to note in the streams.

Also, some edit in the database, as it seems that 127 chars to store id is not enough (-> 255)

_Note: there is no management of attachments when creating post locally (yet)._
_**Note: upgrade to 0.2.0 is needed to apply the changes to the database**_


lib/Migration/Version0002Date20190118124201.php
```diff
@@ -0,0 +1,143 @@
...
+
+
+	/** @var IDBConnection */
+	private $connection;
+
+
+	/** @var array */
+	public static $editToChar255 = [
+		[CoreRequestBuilder::TABLE_CACHE_DOCUMENTS, 'id'],
+		[CoreRequestBuilder::TABLE_CACHE_DOCUMENTS, 'url'],
```
juliushaertl: @daita char 255 will probably be not enought for urls, 2000 chars is a common limit that is used by most modern browsers.

juliushaertl: Same probably also goes for every field where an activity pub url is used as id.

daita: switching all to 2000

juliushaertl:
Let's do this :rocket:

PR #374, opened 2019-01-24
[huge] Announce object and caching during incoming request

daita:


daita:
Technical details: 
- Announce is the way to RT, or boost, a Note.
- An _Announce_ will contains the Id of the boosted object. This item needs to be put in the cache if not yet available in our local database.

Some huge modifications in this PR to be noted:
- Note extends Model/Stream, as the new object Announce. (in the PR #282, Internal/SocialAppNotificationInterface will need to do the same). This way, we will send in the stream different _type_ of objects (Note, Notifications, Announce, ...)
- Incoming request will have a token assigned that will be used in the async process.
- Every incoming request will be async and some check/stuff will be done after closing the incoming connection.
- Those check/stuff will also be done in a background process (retry-on-fail)

More technical details:
- The check/stuff are mainly to check if there is stuff that needed to be cached.
- The cache will be stored as an entry in the stream database (server_notes that should be renamed into server_stream). The content will also be stored in the Announce object to avoid huge [left joined] SQL request.
- This 'layer' can be used to fill a specific use case where a remote instance of ActivityPub reply to a Note from a 3rd instance: if the instance is not available during the verification of the signature, we can assign a background process to verify the signature later.

daita:
on the screenshot, there is _actor_info_ ! 

I will have a look right now

daita:
@jancborchardt @juliushaertl  I will name the action 'Boost' like in mastodon, do we have another proposition ?

jancborchardt:
@daita yes, "Boost" is good! I was thinking about "Repeat" as well, but it looks too much like "Reply" and is a bit boring. **Boost** is much better! :)

daita:
Fixing an issue reported by @juliushaertl we manage boosts, but we do not allow to create them.
Last push should fix this, including some edit in the core to limit a search of a Note based on the current user. 

The generation can be done by command line:

>     ./occ social:note:boost user_id post_id

example: `./occ social:note:boost cult "https://test.artificial-owl.com/apps/social/@test1/15465083831632326429"`

or with a POST request on [/api/v1/post/boost](https://github.com/nextcloud/social/blob/3c1def1d1c5a64ebeeabcd46893aae6fc602c1ab/appinfo/routes.php#L74) that link to [LocalController::postBoost](https://github.com/nextcloud/social/blob/3c1def1d1c5a64ebeeabcd46893aae6fc602c1ab/lib/Controller/LocalController.php#L208-L223)

lib/Migration/Version0002Date20190121153145.php
```diff
@@ -0,0 +1,157 @@
...
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable(CoreRequestBuilder::TABLE_SERVER_NOTES);
+		if (!$table->hasColumn('cache')) {
+			$table->addColumn('cache', Type::TEXT, ['notnull' => false]);
+		}
+
+		$table = $schema->getTable(CoreRequestBuilder::TABLE_SERVER_NOTES);
+		if (!$table->hasColumn('activity_id')) {
+			$table->addColumn('activity_id', Type::STRING, ['notnull' => false, 'length' => 2000]);
```
juliushaertl: Causing an issue on my setup due to the maximum row size:

```
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'ALTER TABLE oc_social_server_notes ADD cache LONGTEXT DEFAULT NULL, ADD activity_id VARCHAR(2000) DEFAULT NULL, ADD object_id VARCHAR(2000) DEFAULT NULL': SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
```

We probably need to move some varchar(2000) fields to text.

daita: switch to 1000 char-length varchar. Please try again after the migrations from #439

aliceinwire:
tested and working

aliceinwire:
but still have problems adding fsf@status.fsf.org
`
500 - {"status":-1,"exception":"OCA\\Social\\Exceptions\\RequestServerException","message":"500 - {\"host\":\"status.fsf.org\",\"url\":\"\/.well-known\/webfinger\",\"timeout\":10,\"type\":0,\"data\":{\"resource\":\"acct:fsf@status.fsf.org\"}}"}`

aliceinwire:
```

  | Error: Call to a member function getPreferredUsername() on null
-- | --


```

daita:
@aliceinwire thanks for the catch, this should be fixed now

aliceinwire:
@daita thanks. I could test and is working!

PR #388, opened 2019-02-01
custom apps folder and webfinger

daita:
This should fix 2 use case:

- the social app is installed in a custom apps folder.
- admin can changes htaccess.RewriteBase after setting up the social app.

### DIFF
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -242,6 +242,7 @@ class ConfigService {
 		return $this->config->getSystemValue($key, '');
 	}
 
+
 	/**
 	 * @param string $cloudAddress
 	 */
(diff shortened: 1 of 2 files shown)


lib/Service/ConfigService.php
```diff
@@ -277,13 +278,17 @@ public function getCloudAddress(bool $host = false) {
...
 	/**
-	 * // TODO - check the Apps folder
+	 * @param string $path
 	 *
 	 * @return string
 	 * @throws SocialAppConfigException
 	 */
-	public function getUrlSocial(): string {
-		return $this->getCloudAddress() . '/apps/social/';
+	public function getUrlSocial(string $path = ''): string {
```
juliushaertl: Path doesn't seem to be used anywhere? Is it actually needed?

juliushaertl: Nevermind, had the wrong branch checked out

PR #398, opened 2019-02-06
Use chunkhash to properly trigger reload of webpack chunks on upgrade

### LINKED ISSUE #373: every page is blank except profile
I've installed social and reset it a few times using occ but every page is blank except for the profile page. nothing in the error logs, it just seems to do nothing.

NC 15.0.2
Social 0.1.3

If a write a post and send it, it seems to go somewhere but it's not visible on any of the pages.
image

juliushaertl:
Fixes #373 

Instead of profile.js as the name for chunks we will now use profile.[chunkhash].js to properly force the browser to reload those files instead of using the ones from the cache.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -7,7 +7,8 @@ module.exports = {
 	output: {
 		path: path.resolve(__dirname, './js'),
 		publicPath: '/js/',
-		filename: 'social.js'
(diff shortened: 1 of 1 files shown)


daita:
looks good to me

PR #404, opened 2019-02-07
Delete Note from db if author is no more in cache

daita:
This is an hotfix for the missing actor_info on the front-end.

We should also remove the Notes when the author is deleted from a remote instance.

### DIFF
--- a/lib/Db/NotesRequest.php
+++ b/lib/Db/NotesRequest.php
@@ -364,16 +364,5 @@ class NotesRequest extends NotesRequestBuilder {
 		return $notes;
 	}
 
-
-	/**
-	 * @param string $id
-	 */
-	public function deleteNoteById(string $id) {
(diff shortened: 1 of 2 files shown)


daita:
I don't like this solution. **This will not be merged.**  (you can still use this branch to delete bogus entries by reloading the app)

I think the front-end should just check if actor_info exists or not, as the data from actor_info is used just to display the right avatar. If no actor_info, no avatar.

The faulty entries will be managed by a background process.

violoncelloCH:
@daita I've done a complete reset of the app and now I can't reproduce the actor_info console error, so I can't test this...

daita:
If you want to reproduce the issue to help you fixing fix it on the front-end: you need to delete an entry from the cache_actors database. The deleted entry needs to be an author of a Note.

juliushaertl:
Replaced by #412

PR #433, opened 2019-02-22
create object even if it can't be signed

daita:


daita:
On friday, users encountered some downtime in creating and receiving new Notes. 

To avoid this, caching.

daita:
Caching the content of context url: 

- if file is in appdata, we use it,
- if file is not in appdata, we download it,
- if download is not effective, we get the file from the context/ folder locally.

juliushaertl:
What is the reason we need to dowload them if we have stored them in the repo? Or can they be different on different instances/change later?

daita:
> What is the reason we need to dowload them if we have stored them in the repo? Or can they be different on different instances/change later?

no real reason, this is why it is only downloaded if the file is not in appdata. But still downloadable in case it's not the same version than the one in the app itself.

Issue #438, opened 2019-02-24
fosstodon account add fail


aliceinwire:
```

  | 500  -  {"status":-1,"exception":"OCA\\Social\\Exceptions\\InvalidResourceException","message":"An  exception occurred while executing 'INSERT INTO  `social_cache_documents` (`id`, `type`, `url`, `media_type`,  `mime_type`, `error`, `local_copy`, `public`, `creation`) VALUES(?, ?,  ?, ?, ?, ?, ?, ?, ?)' with params  [\"https:\\\/\\\/cloud.example.com\\\/documents\\\/g\\\/q4p5431e-1t34-1tf4-812c-8738b2839c93\",  \"Image\",  \"https:\\\/\\\/cf.mastohost.com\\\/v1\\\/AUTH_XXXXX\\\/fosstodon\\\/accounts\\\/avatars\\\/000\\\/088\\\/546\\\/original\\\/9ac237b107442c93.jpg\",  \"image\\\/jpeg\", \"\", 0, \"\", \"0\", \"2019-02-24  08:21:52\"]:\n\nSQLSTATE[22001]: String data, right truncated: 1406 Data  too long for column 'url' at row 1"}
-- | --

```

aliceinwire:
resetting didn't help.
I'm on 0.2.0-dev01

also this error:

xt | 500 - {"status":-1,"exception":"OCA\\Social\\Exceptions\\ItemUnknownException","message":""}
-- | --

aliceinwire:
I could solve my issue by getting into mysql and doing a horrible hack :sob: 
```
use nextcloud;
ALTER TABLE social_cache_documents modify url varchar(500);
```

have fun!

aliceinwire:
Probably we could write a migration for it.

daita:
thanks for the catch, i'll work on some migrations !

violoncelloCH:
fixed by #439

PR #439, opened 2019-02-26
Enlarge some fields

daita:
#438

### DIFF
--- a/lib/Migration/Version0002Date20190226000001.php
+++ b/lib/Migration/Version0002Date20190226000001.php
@@ -55,7 +55,7 @@ class Version0002Date20190226000001 extends SimpleMigrationStep {
 
 
 	/** @var array */
-	public static $editToChar2000 = [
+	public static $editToChar1000 = [
 		[CoreRequestBuilder::TABLE_CACHE_ACTORS, 'id'],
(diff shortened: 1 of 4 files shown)


lib/Migration/Version0002Date20190226000001.php
```diff
@@ -0,0 +1,154 @@
...
+ */
+class Version0002Date20190226000001 extends SimpleMigrationStep {
+
+
+	/** @var IDBConnection */
+	private $connection;
+
+
+	/** @var array */
+	public static $editToChar2000 = [
```
juliushaertl: Same as in https://github.com/nextcloud/social/pull/374#discussion_r260458387

daita:
@aliceinwire because you're not using innodb :)

@juliushaertl can you check now ?

juliushaertl:
@daita Nope still an issue 

```
 An exception occurred while executing 'ALTER TABLE oc_social_server_notes ADD id VARCHAR(1000) DEFAULT NULL, ADD   
  `to` VARCHAR(1000) DEFAULT NULL, ADD attributed_to VARCHAR(1000) DEFAULT NULL, ADD in_reply_to VARCHAR(1000) DEFA  
  ULT NULL':  
```

PR #453, opened 2019-03-05
creating a new field id_prim, set as primary and fill it

daita:
Issue is that length of field to be set as primary are limited and _might_ be an issue on the fediverse. Solution from @blizzz  is to hash the string id to an hash, and stored the hash in the primary key.

It might slow done some requests, but we could also improve those request by using the generated primary key 'id_prim' instead of 'id'

### DIFF
--- a/lib/Migration/Version0002Date20190305091901.php
+++ b/lib/Migration/Version0002Date20190305091901.php
@@ -130,8 +130,6 @@ class Version0002Date20190305091901 extends SimpleMigrationStep {
 					   ->eq($field, $update->createNamedParameter($id))
 				);
 				$update->execute();
-
-				$update->execute();
 			}
 			$cursor->closeCursor();
 
--- a/lib/Migration/Version0002Date20190305091901.php
+++ b/lib/Migration/Version0002Date20190305091901.php
@@ -1,0 +1,131 @@
+<?php
+declare(strict_types=1);
+
+
+/**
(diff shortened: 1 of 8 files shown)


lib/Migration/Version0002Date20190305091901.php
```diff
@@ -0,0 +1,131 @@
...
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
+
+		foreach (self::$setAsKeys as $edit) {
+			list($tableName, $field) = $edit;
+
+			$prim = $this->getPrimField($field);
+			$qb = $this->connection->getQueryBuilder();
+			$qb->update($tableName)
+			   ->set($prim, $qb->createFunction('SHA2(' . $field . ', 512)'))
```
juliushaertl: SHA2 is not available on sqlite

lib/Migration/Version0002Date20190305091901.php
```diff
@@ -0,0 +1,152 @@
...
+				$update = $this->connection->getQueryBuilder();
+				$update->update($tableName);
+				$update->set($prim, $update->createNamedParameter($hash));
+				$update->where(
+					$qb->expr()
+					   ->eq($field, $update->createNamedParameter($id))
+				);
+				$update->execute();
+
+				$update->execute();
```
juliushaertl: Duplicate execute call

PR #462, opened 2019-03-12
boost button

violoncelloCH:
* Target version: master 

### Summary
- [X] button for boosting (reposting)
- [x] indicate if a post is boosted and make it possible to unboost


violoncelloCH:
@daita cool! I'll hopefully find the time to look into this next week :)

how should an Announce in the response look like? I only get this:
image
Could the Announce directly contain the original note in the same format as it was a note, just with the additional info that it's a boost and by whom? If so, the front end part for showing it's a boost would be pretty simple to implement...

violoncelloCH:
Of course we can split this up into two separate PRs... (I can do so once I'll continue working on this)


> Note that a cache version of the original boosted Post should be available in the Announce object. Don't know why you don't have it in your own stream

I'll probably completely reset social on my development environment and see if this helps...

daita:
Yes, might be better to reset your current setup :]

Also, the back-end part of the PR should work right now, the original object should be available in the 'cache' entry:

Selection_025

- cache._items is used by the caching system.
- cache.\<url\>.content contains the original Post with id=\<url\>
- cache.\<url\>.status _equals_ 9 when the cache is ok, if not equals to 9 just write that the Post is not available.

Value of status: https://github.com/nextcloud/social/blob/accf7e424d23e8f343f4d0ecf053f65ab2800e77/lib/Model/StreamQueue.php#L53-L55

daita:
@juliushaertl can you have a look to the back-end ? and maybe provide some help to @violoncelloCH to finalize the front-end.

daita:
@violoncelloCH do you need anything to go on with the front-end part ?

violoncelloCH:
@jancborchardt how should we highlight the boost icon/button if the post was boosted?
I would have done it by changing the icon color to the primary color, however @daita mentioned that in his case this would be black; so no difference... should we do this somehow with a background?

jancborchardt:
@violoncelloCH good point – I would put a filled blue circle (not themed, just like stars are always yellow / hearts are always red) around it and then make the icon white – which would also work as a hover/focus effect. What do you think?

violoncelloCH:
@jancborchardt so I've implemented it this way with a blue background:
image
when you hover it, the blue gets a little bit darker... what do you think?

juliushaertl:
> is there a general point, where the stream elements data is parsed and I could do a general check if item.action.values.boosted is defined and if not set it to false? @daita said the backend simply skips the whole action part if there was never a action set on a Post, so the frontend should simply consider boosted as false. What I'm now doing is checking in several places if action is undefined what definitely does not make sense... what would be the correct place to check this globally?

There is no parsing in general, we just copy the json data over to the vuex store. So the way to solve this would be to just have the logic once in the computed property. Then you can reuse that in the boost() method as well

```
		isBoosted() {
			if (typeof this.item.action === 'undefined') {
				return false;
			}

			return !!this.item.action.values.boosted;
		}

// ...

		boost() {
			if (this.isBoosted) {
				this.$store.dispatch('postUnBoost', this.item)
			} else {
				this.$store.dispatch('postBoost', this.item)
			}
		}
```

jancborchardt:
@violoncelloCH hmm not sure, as then when it’s active it looks like a button. Might be a bit much with the circle around it. Let’s instead just make it opaque when retweeted so it’s darker, and also make it theming/primary color blue.

violoncelloCH:
thank you very much @juliushaertl 
fixed that in 30282f7

@jancborchardt thanks for the feedback! I get your point
would you use "nextcloud blue" as a static color value or `$color-primary` now?
image

jancborchardt:
@violoncelloch right, let's do Nextcloud blue static color value so it's the same for everyone. :)

daita:
also Alpha, this can be improved in a small PR later !

violoncelloCH:
implemented this :)
so I think we are ready for review?

jancborchardt:
@violoncelloCH it looks good as said! :) Probably it’s only something for a future pull request – we could think about an animation kind of like Twitter does for liking and retweeting. :tada:

lib/Controller/NavigationController.php
```diff
@@ -189,15 +189,15 @@ public function navigate(string $path = ''): TemplateResponse {
...
+//		$frontControllerActive =
+//			($this->config->getSystemValue('htaccess.IgnoreFrontController', false) === true
+//			 || getenv('front_controller_active') === 'true');
 
 		$cloudAddress = rtrim($this->config->getSystemValue('overwrite.cli.url', ''), '/');
 		if ($cloudAddress !== '') {
-			if (!$frontControllerActive) {
-				$cloudAddress .= '/index.php';
-			}
+//			if (!$frontControllerActive) {
```
juliushaertl: Why is this commented out? This will not work on instances where the rewrite is not set.

violoncelloCH:
@jancborchardt what I could do without much effort would be a simple rotation of the button with an `css @keyframes` animation... however this animation would also be triggered on loading the page (if there are boosted posts)... would you do this? something more complex is probably better something for a future PR... It's more important to implement the displaying of boosted posts now that we have boosting itself...

juliushaertl:
> however this animation would also be triggered on loading the page (if there are boosted posts).

You can also just set the animation to a custom class like `.boost-animated`. If you apply the class with vue js later on, the animation will only run once triggered.

> something more complex is probably better something for a future PR

:+1: Yes, let's not grow this one even bigger :wink:

daita:
merging despite some strange entries in the console log.

PR #507, opened 2019-05-04
Bugfix/boost

juliushaertl:
@daita This should fix the console errors

Note that https://github.com/nextcloud/social/commit/cc402ada48831e0ae4e4a24eabebcbf39ae7360b will make sure that timeline entries with an empty content attribute (boosts) are always shown, but they have no content.

image

Should those be hidden by default or is it a bug that the content is missing?

### DIFF
--- a/lib/Db/NotesRequest.php
+++ b/lib/Db/NotesRequest.php
@@ -155,7 +155,7 @@ class NotesRequest extends NotesRequestBuilder {
 
 		$qb = $this->getNotesSelectSql();
 		$this->limitToIdString($qb, $id);
-//		$this->limitToType($qb, Note::TYPE);
+		$this->leftJoinCacheActors($qb, 'attributed_to');
 
 		if ($asViewer) {
 			$this->limitToViewer($qb);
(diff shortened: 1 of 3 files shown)


juliushaertl:
- [ ] We need to check the cache entry for the boosted post data to properly show its content and the original author of the post:

image

daita:
@juliushaertl last commit should do the trick

daita:
ok, so actor_info is well saved in the cache version of the boosted post.

But the boost does not display anything. (until I rewrite the migration of the database, let's assume that a user can see his own Boost as an Announce, not just as a Post with an 'action')

However, during the process of boost/unboost, I have this error:
```
[Vue warn]: Error in callback for watcher "function () { return this._data.$$state }": "Error: [vuex] do not mutate vuex store state outside mutation handlers."

(found in <Root>)
```

Also, a use-case to be filtered:

- user1 create a new Post,
- user2 boost the Post from user1,
- user1 can boost the boost from user2 of his own post!

We should not allow this (remove the icon on the front-end), I will make the back-end returns an error.

cc @violoncelloCH

juliushaertl:
> [Vue warn]: Error in callback for watcher "function () { return this._data.$$state }": "Error: [vuex] do not mutate vuex store state outside mutation handlers."

@daita This is actually what the first commit in this PR fixes.

daita:
ok, surely some conflict with another branch, had to remove all file and deploy again and it is now working :]

Issue #324, opened 2019-01-10
Show reposts (aka retoots) in personal timeline


Scumpeter:
**Is your feature request related to a problem? Please describe.**
When you follow an account that reposts (retoots) something, it is currentliy (social 0.1.2) not shown in your timeline.

**Describe the solution you'd like**
When a person reposts something, it should be visible in any timeline where the posts from this person are visible.
Reposts should be marked as reposts and include information about the original poster.

**Describe alternatives you've considered**
none

**Additional context**
The behaviour should be similar to that in Twitter or Mastodon.

PR #546, opened 2019-05-28
display boosted posts with an according indication

violoncelloCH:
* Resolves: #324 
* Target version: master 

### Summary
This implements the displaying of reposts (boosted / retooted posts) and displays who has boosted it in a short indication:
image

cc @jancborchardt for design feedback (because I'm not sure what the best way to display this boost message is... maybe right-align the icon and vertically align it to the centre of the text?)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -12,12 +12,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/daita/my-small-php-tools.git",
-                "reference": "732d54bca742e3ecdb2b544589550a37172c1258"
+                "reference": "6e8f346a2ee488655316d1e4139c27417d6b7e4d"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/daita/my-small-php-tools/zipball/732d54bca742e3ecdb2b544589550a37172c1258",
-                "reference": "732d54bca742e3ecdb2b544589550a37172c1258",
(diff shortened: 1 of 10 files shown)


jancborchardt:
@violoncelloCH looks good! Just 3 details:
- The `username boosted` text could be grey / half-transparent for less focus on that
- The username of the boost doesn’t need to be bold
- The boost icon should move so it’s vertically aligned with the text and right-aligned with the avatar below (see this screenshot):
boosting

daita:
so, I made a test with:

- userA create a new Post (ok)
- userB boost that new Post (ok)
- userC boost the boosted Post. (fail)

On boosting the boosted post, the userC got an error. The link to boost is using the PostId of the Announce, not the Original Post.

violoncelloCH:
@jancborchardt thanks for the feedback! I'll implement your proposal!

@daita are you sure? `TimelineContent.vue`, which does the boosting, should access the cached object and therefore take the id of the cached Note (as long as the backend sends the correct id inside of the cached Note)... Can you double check if it's not the `Type error: ...action is undefined` I'm already working on? (in this case the Note is boosted, but the frontend displays an error, because it can't set the boosted state)

daita:
@violoncelloCH just checked, the postId is an Announce, not a Post. Also, if I click on 'Reply-To' the text 
'In reply to <author of the Boost>' (instead of <author of the Post>) is available above the textarea. Maybe adding both authors (Boost+Post) to the Reply-To list.

Issue #560, opened 2019-06-01
new lines are trimmed from posts


guzzisti:
**Describe the bug**
Newlines are completely trimed from posts, no matter if remote posts (e.g. Mastodon) or a post in Nextcloud Social.

**To Reproduce**
1. Create a post that contains newlines
2. Click on "Post"
3. Post is shown without newlines.

**Expected behavior**
Newlines are honored in posts.

**Screenshots**
![image](https://user-images.githubusercontent.com/10488791/58747469-11b63900-846c-11e9-99e8-72a58291c5f8.png)

Original post text was:
```
Das ist ein Test-Post

- mit - 

Zeilenumbrüchen.
```

**Social app version:**
0.2.1
**Operating system**:
Manjaro Linux
**Web server:**
Apache
**Database:**
MySQL 5.7
**PHP version:**
7.3
**Nextcloud version:** (see Nextcloud admin page)
16.0

guzzisti:
I tried to digg a bit deeper into the code, for me it seems like newlines are encoded as html entities and these are stripped completely of the post text:
![Screenshot_2019-06-01_13-23-28](https://user-images.githubusercontent.com/10488791/58747819-b76ba700-8470-11e9-8c15-1241a2792229.png)

daita:
@juliushaertl @violoncelloCH  any idea about this ? I will check on my side if the back-end is doing something bad.

guzzisti:
I guess it's a backend issue as it happens to federated posts as well, e.g. Mastodon:
original
![image](https://user-images.githubusercontent.com/10488791/58761328-53fe7980-8543-11e9-8338-09713af22c97.png)

Nextcloud Social
![image](https://user-images.githubusercontent.com/10488791/58761330-5fea3b80-8543-11e9-8aab-4c8fd1f0ffba.png)

daita:
could be the front-end :-)

I checked and the content sent by the front-end to the back-end when creating a new Post is already parsed/trimed

violoncelloCH:
it could also be that the frontend trims it on displaying as well...
I can check this, but not sure when, because I currently don't have much spare time...

PR #597, opened 2019-06-25
🐛 FIX: line breaks not handled correctly

violoncelloCH:
* Resolves: #560 
* Target version: master 

### Summary
this should remove all html attributes except line breaks, but replaces them with `\n` before sending it to the backend and it also properly makes `<br>`s out of `\n`s again when displaying a post


StCyr:
rebased but still not correct

StCyr:
should be good now @daita (cc: @violoncelloCH)

violoncelloCH:
thanks for finishing this one @StCyr !

Issue #347, opened 2019-01-17
Roadmap/Goals?


KopfKrieg:
I hope I didn't miss it, but is there some kind of roadmap or a list of goals for the Social app? If not, I'd like to open some kind of discussion about the topic, because I've got a few questions.

As far as I know the Social app implements the ActivityPub protocol which means - in theory - it could work as an (more or less complete) replacement for, e.g., Mastodon (when I say replacement I mean I don't have to setup my own Mastodon instance because I could just use Nextcloud with the Social app).

This would also open up the possibility to allow clients which speak the ActivityPub protocol (e.g. Tusky, Mastalab, Twidere, AndStatus, etc.) to connect to a Nextcloud instance (which would be awesome).

Or was the idea behind the Social app to be used between Nextcloud users and support for other Instances through the ActivityPub protocol (e.g. Mastodon) was just a coincidence? On the other hand, that wouldn't really make sense, because we have Nextcloud Talk and Circles for that, right?

**Edit**: I just found issue #145. So, I guess it could really act as a replacement for standalone Mastodon servers?

jospoortvliet:
I can recommend to look at the milestones: https://github.com/nextcloud/social/milestones

That's essentially the current roadmap.

it is indeed true that Nc is a ActivityPup server able to connect to Mastodon servers and federate with others who have implemented this API.

PR #604, opened 2019-07-03
Update README.md

jospoortvliet:
* Resolves: #347
* Target version: master 

### Summary
Re-arranged to put features on top; put in link to announcement blog and milestones

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,6 @@ The app is in alpha stage, so it’s time for you to [get involved! 👩‍💻]
 For a roadmap, see our [development milestones](https://github.com/nextcloud/social/milestones) - feel free to pick up any of the tasks which are not yet in progress!
 
 Some requirements in this alpha stage are that your Nextcloud:
-- must use a real SSL certificate
-- must be accessible from the internet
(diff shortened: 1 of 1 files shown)


jancborchardt:
Yes, much better to put the details further down :+1:

README.md
```diff
@@ -17,8 +17,6 @@ The app is in alpha stage, so it’s time for you to [get involved! 👩‍💻]
 For a roadmap, see our [development milestones](https://github.com/nextcloud/social/milestones) - feel free to pick up any of the tasks which are not yet in progress!
 
 Some requirements in this alpha stage are that your Nextcloud:
-- must use a real SSL certificate
-- must be accessible from the internet
 - must run on the default port
```
jancborchardt: @daita @violoncelloCH @juliushaertl so can we get rid of this default port requirement? Or is this the case for the rest of Nextcloud anyway? Or is it a non-issue because no one uses other ports in the first place?

We can always just remove it and then if someone opens an issue look into it?

daita: Yes, we can get rid of it

PR #618, opened 2019-07-08
parse as content

daita:
back-end will keep `<br>` tag. 
We need front-end to treat it like HTML.

### DIFF
--- a/lib/Model/ActivityPub/ACore.php
+++ b/lib/Model/ActivityPub/ACore.php
@@ -61,6 +61,7 @@ class ACore extends Item implements JsonSerializable {
 	const AS_USERNAME = 5;
 	const AS_ACCOUNT = 6;
 	const AS_STRING = 7;
+	const AS_CONTENT = 8;
 	const AS_TAGS = 10;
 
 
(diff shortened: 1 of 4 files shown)


violoncelloCH:
so do I get it right, that this supersedes #597? 
should I add a commit here which makes the frontend just displaying the `message` the way it comes from the backend? trusting all html in it is safe to be put to the DOM?

daita:
Yes, would be nice (and the missing part of this PR !)

violoncelloCH:
okay that's a little bit more difficult, because actually the linkify dependency we're using is removing all the html tags... I'll see if I find a way to tell it not to do so...

daita:
Yes, i tried to reach @juliushaertl about it, if we could set a list of allowed tags

violoncelloCH:
@daita with this backend, is the frontend supposed to send linebreaks as `<br>` html tags? because when I do a request for creating a new post having `<br>`s in the message content, they still don't appear again in the stream request again...

@juliushaertl I don't (yet) really get how this input form is working:
firstly: why does it put every line into a new `<div>` container?
then: if I do a line break directly at the end of a word, it just creates a new `<div>` for the second line... only if have a space `" "` at the end of the line, it inserts also a `<br>`; however at the end of the message, there is always a `<br>`, no matter if we have line breaks at all or not

juliushaertl:
@daita If we can trust the backend we could use https://soapbox.github.io/linkifyjs/docs/linkify-html.html which doesn't strip the html then. But we should probably still sanitize the html after the linkify run then in the frontend as well.

> I don't (yet) really get how this input form is working:
> firstly: why does it put every line into a new <div> container?
> then: if I do a line break directly at the end of a word, it just creates a new <div> for the second line... only if have a space " " at the end of the line, it inserts also a <br>; however at the end of the message, there is always a <br>, no matter if we have line breaks at all or not

This is actually the behavior of the browser in a contenteditable div. I'm not sure yet how we could solve this properly

daita:
we can trust the content sent by the back-end. I only let BR doing throw

daita:
Currently, the back-end does not receive any html within the content

daita:
already fixed

PR #624, opened 2019-07-11
bypass $as

daita:


StCyr:
yes! your change works: the attachements are now received on the front-end side 👍

StCyr:
hmmm, some posts are missing their attachements though :-(

daita:
check the table `oc_social_a2_cache_documts` and the value of the field `local_copy`. if empty, it means the cached version is not available

StCyr:
> check the table `oc_social_a2_cache_documts` and the value of the field `local_copy`. if empty, it means the cached version is not available

I don't understand: A cached version of? 

BTW, I'd also changed line: https://github.com/nextcloud/social/blob/ca6e1e6c58078999ed4789ad05e6351e6814a1bb/lib/Model/ActivityPub/Object/Note.php#L131 

to:

```
		$this->importAttachments($this->getArray('attachments', $data, []));
```

on my server (note the additional 's').

I believe it should be changed here also.

daita:
looks like there is some issue with the importation of attachments. I'll have a look

daita:
well, one of the issue was to edit the `Note.php` file :)

`'attachment'` with no `'s'` is fine, why would you change this in the `import()` method ?

StCyr:
Because there’s a ‘s’ at line 184

But, probably I don’t understand the code enough. I’ll rollback the change I made on my prod server here then.

Thanks,

Cyrille

daita:
`import()` is called on incoming ActivityPub request, so 'attachment' is used by the protocol.
`importFromDatabase()` is called on retreiving data from database, and the field entry is 'attachments'

daita:
ok, my bad, pushed everything on `master` instead of your branch `showAttachments`, we're close to the week-end :-)

Please rebase your branch on master and check if attachments are working better. Also, if you miss some attachments, try: `./occ social:cache:refresh`

PR #631, opened 2019-07-13
Adds ILogger to several classes

StCyr:
* Resolves: None
* Target version: master 

### Summary

A man needs logging


daita:
I usually use the alias log() within MiscService

StCyr:
ah! good to know :-)

I guess we can close this PR then :-p

Issue #640, opened 2019-07-18
New route to get all liked posts


StCyr:
I've implemented recently the front-end for the 'like' functionality, and I think it would make sense to have a navigation menu entry to show all posts ever liked.

I think it first requires a new route to in the backend.

daita:
should be `/api/v1/stream/liked`

Please note that all route for the (local) API will be changed: `/api/v1/` will be changed to `/local/v1/` we will keep /api/v1 available for mastodon client API

StCyr:
This is solved by #642

PR #642, opened 2019-07-19
Add 'liked' stream

daita:


daita:
back-end part for #640

PR #643, opened 2019-07-20
St cyr fix640

StCyr:
* Resolves: #640
* Target version: master 

### Summary

Adds a menu entry to show a "liked" timeline


daita:
nice feature, simple implementation.

Would be nice to remove the message from the list on unlike ?

Issue #106, opened 2018-12-03
Adding images to posts


jancborchardt:
People should be able to upload images to their posts. Just like we have in the Talk app, with a ➕ button (or 🖼) which then shows you:
- ⬆ Upload from device
- 📁 Add from Nextcloud

PR #657, opened 2019-07-23
upload post attachments

StCyr:
* Resolves: #106 
* Target version: master 

### Summary

Makes the Composer able to upload attachments.

This is a work in progress

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -7,7 +7,7 @@
   @include icon-color('boost', 'social', '#0082c9', 1);
 }
 .icon-upload {
-  @include icon-color('upload', 'actions', $color-white, 1, true);
+  @include icon-color('upload', 'actions', $color-black, 1, true);
 }
 
 img.emoji {
--- a/css/style.scss
+++ b/css/style.scss
@@ -1,9 +1,13 @@
 @include icon-black-white('reply', 'social', 1);
 @include icon-black-white('emoji', 'social', 1);
 @include icon-black-white('boost', 'social', 1);
+@include icon-black-white('upload', 'actions', 1, true);
 
 .icon-boosted {
   @include icon-color('boost', 'social', '#0082c9', 1);
+}
+.icon-upload {
+  @include icon-color('upload', 'actions', $color-white, 1, true);
 }
 
 img.emoji {
--- a/lib/Controller/LocalController.php
+++ b/lib/Controller/LocalController.php
@@ -48,9 +48,9 @@ use OCA\Social\Service\FollowService;
 use OCA\Social\Service\HashtagService;
 use OCA\Social\Service\LikeService;
 use OCA\Social\Service\MiscService;
-use OCA\Social\Service\StreamService;
 use OCA\Social\Service\PostService;
 use OCA\Social\Service\SearchService;
+use OCA\Social\Service\StreamService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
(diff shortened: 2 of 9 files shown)


StCyr:
The 3 first commits make it possible to upload files and display them resized under the vue-tribute component of the Composer.

It's currently quite ugly but it works.

daita:
@nickvergessen I now remember why I wanted to talk with you earlier !

What do you think about this ? we're looking for a way to attach file/document to a new Post. What do you use in the Talk app ?

nickvergessen:
The plan is documented in https://github.com/nextcloud/spreed/issues/1811#issuecomment-505096933

Whether it works in an okay-fashion we will see in the 18 cycle when we try to work on it.

daita:
the original image is sent to the backend at post creation, i'll work on the caching

StCyr:
This is taking shape 👍

daita:
This is nice, really good job !


post_attachments



I'll do some more testing tomorrow on my side.
Some issues I have encountered:

- [ ] User should be able to remove a document from the list
- What happens if the document is not an image ?
- [ ] It seems that the gallery does not disappears after the creation of the Post.
- [ ] ugly button

Can you please rebase on master ?

StCyr:
at 2e56fef should be:

- [X] User should be able to remove a document from the list
=> for the moment, just clicking on the miniature is enough to remove it. I would like to rather add a small (X) cross icon in the upper right corner of all miniatures and that only clicking on that cross icon would remove the document
- What happens if the document is not an image ?
=> For the moment an error will most probably pop up. We should handle the case and maybe generate a sort of general purpose document miniature with the name of the attachment
- [X] It seems that the gallery does not disappears after the creation of the Post.
- [ ] ugly button

==========

daita:
>     * I would like to rather add a small (X) cross icon in the upper right corner of all miniatures and that only clicking on that cross icon would remove the document

Exactly what I was thinking about, but at least with have the functionality working now.
I tested and it's nice.Iit only works for image (in the back-end) but this is already good enough !

jancborchardt:
> the one to upload
> the X icon to delete an attached file during post creation

`.icon-upload` is good for upload, and `.icon-close` is good for removing the attached file. Make sure you use the delete icon in a button also for proper accessibility. :)

StCyr:
I use icon-upload for the upload button but it lacks a nice circle if I don’t add a « border-radius: 50% » style.

Is it normal?

I would like it to have a nice circular border as the « change post audience » button just beside (the « earth » button.

Cyrille

jancborchardt:
@StCyr riight, it’s because it is `input type="file"`, and that is not defined in the global styles. If you like you can modify the `core/css/inputs.scss` in the server repository?

Also, images could have a slight border around them, use the var(--color-border) variable, and also a var(--border-radius) to have some container around them to know what it is.

StCyr:
What do you think about the ways it looks now @daita ?

StCyr:
Yes, it should be mergeable as is.

As you said some additional eye-candies would be nice but I don't want to spend more time in these atm.

Cyrille

==========

For the record, here are some possible look&feel improvements:

1- Only show the 'close' badge when hovering on an image;
2- Nicer generic icon for non image file types.

Other worthwile improvements:

1- Back-end should support non image files;
2- Support attaching several files at once.

jancborchardt:
> 1- Only show the 'close' badge when hovering on an image;

It’s ok to show this all the time, so it also works on touch interfaces. :)

PR #666, opened 2019-07-29
Identify hashtags from what's found in a toot's DB entry

StCyr:
* Resolves: 
* Target version: master 

### Summary

Originaly, hashtags were identified in the front-end using linkifyjs. But, this method, amongst other limitations, doesn't support hashtags with non-latin characters.

I then changed to hashtag-regex which brought some improvements. 

Unfortunately, this method still has the problem that if a hashtag is directly followed by some text, social's front-end will wrongly identify the hashtag as in the following toot: "Hello #fediverseHow are you?" (social identifies a "#fediverseHow" hashtag whereas the intended hashtag is "#fediverse" only).
 
The thing is: the list of hashtags in a toot is sent together with the toot (in its 'tag' attribute) and is stored in the DB (field 'hashtags' in the ac_social_a2_stream table).

This PR makes use of this 'hashtags' field to identify hastags.

### Remaining issues

This PR doesn't work better than hastag-regex atm because the 'hashtags' field in the DB doesn't support non-latin characters.

Here's a sample output:

image

We can see that the hashtag isn't fully identified.

And, here's what the DB says about this toot:

```sql
MariaDB [nextcloud]> select * from oc_social_a2_stream where hashtags='["gTest"]'\G;
*************************** 1. row ***************************
           id_prim: 4b02b8ab1458ee096d7f8449eb5febb9a7053983faac84e5671e3a73ba5e226dfaffe99e37356da1943421ab48d5bf359838236be861853f2378c8531923c092
                id: https://nextcloud.bollu.be/index.php/apps/social/@cyrille/15644098951701989652
              type: Note
                to: https://nextcloud.bollu.be/index.php/apps/social/@cyrille/followers
          to_array: []
                cc: ["https://www.w3.org/ns/activitystreams#Public"]
               bcc: []
           content: this is just a test of a toot with a hashtag containing non-latin characters. #gTest&eacute;CeTruc
           summary:
         published: 2019-07-29T14:18:15+00:00
    published_time: 2019-07-29 14:18:15
     attributed_to: https://nextcloud.bollu.be/index.php/apps/social/@cyrille
       in_reply_to:
       activity_id:
         object_id:
          hashtags: ["gTest"]
            source:
         instances: [{"uri":"https://nextcloud.bollu.be/index.php/apps/social/@cyrille/followers","type":3,"priority":1}]
       attachments: []
             cache: []
          creation: 2019-07-29 14:18:16
             local: 1
hidden_on_timeline: 0
           details: []
           subtype:
1 row in set (0.001 sec)

ERROR: No query specified

MariaDB [nextcloud]>

```

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -24,6 +24,7 @@
 	<author mail="jus@bitgrid.net">Julius Härtl</author>
 	<author mail="jonas@violoncello.ch" homepage="https://violoncello.ch">Jonas Sulzer</author>
 	<author mail="hey@jancborchardt.net" homepage="https://jancborchardt.net">Jan-Christoph Borchardt</author>
+	<author mail="cyrpub@bollu.be">Cyrille Bollu</author>
 	<namespace>Social</namespace>
 	<category>social</category>
 	<website>https://github.com/nextcloud/social</website>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -96,6 +96,7 @@ return [
 		['name' => 'Local#globalActorInfo', 'url' => '/api/v1/global/actor/info', 'verb' => 'GET'],
 		['name' => 'Local#globalActorAvatar', 'url' => '/api/v1/global/actor/avatar', 'verb' => 'GET'],
 		['name' => 'Local#globalAccountsSearch', 'url' => '/api/v1/global/accounts/search', 'verb' => 'GET'],
+		['name' => 'Local#globalTagsSearch', 'url' => '/api/v1/global/tags/search', 'verb' => 'GET'],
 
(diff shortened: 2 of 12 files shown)


StCyr:
so, after some discussion with @daita it appears that the hashtags' DB field *does* support non-latin characters, and that the problem lies in fact in the composer.

That's: only local toots have wrong hashtags  (ie: "#HelloF" instead of "#HelloFédiverse"), remote toots have correct hashtags.

I'm now working on the composer to try and improve hashtag detection there.

StCyr:
PR looks good to me now :-)

daita:
small glitch on empty string search; otherwise, I like it !

PR #678, opened 2019-08-12
push service

daita:
_nc18 + https://github.com/daita/stratos_

- renaming _NoteService_ to _StreamService_
- on new stream item, checking viewers'
- generation of events


daita:
(nc18)
Also require https://github.com/daita/stratos

StCyr:
nice :-)

do you want some testing?

PR #687, opened 2019-08-19
Better layout of TimelinePost component

StCyr:
Signed-off-by: Cyrille Bollu <cyrpub@bollu.be>

* Resolves: Not registered
* Target version: master 

### Summary

Moves post's timestamp above its message and attachments so that its message has more horizontal space.

### Screenshot

image


daita:
@nextcloud/designers  : this is a glitch-fixing. Merging !

Issue #686, opened 2019-08-19
empty list of followers 


StCyr:
**Describe the bug**

When browsing to https://fqhn/index.php/apps/social/<my_uid>/followers nothing is shown

**Expected behavior**

My followers' list should be shown

**Screenshots**

![image](https://user-images.githubusercontent.com/8179031/63249706-4915d400-c26a-11e9-8909-4eef7b9dc23a.png)

**Possible cause**

I believe this issue is due to `ActivityPubController`'s `followers` method not returning `actor_info` (contrary to its `following` method)

daita:
the table `social_follows` need some purge when an account is deleted

StCyr:
indeed, a `delete from oc_social_a2_follows where actor_id like 'https://mstdn.jp%';` solved my issue

StCyr:
I think other things might need to be deleted because the front-end still shows 6 followers while I've deleted 3 (Yes I know I'm popular :-p )

daita:
do you have a cron running on your Nextcloud ? if not, it might the issue.

Please try : `./occ social:cache:refresh`

daita:
closing this one ?

PR #691, opened 2019-08-21
Cleaning followers on deleted accounts

daita:


daita:
Should fix #686

(Not tested yet, careful)

Issue #690, opened 2019-08-21
Profile view doesn't show properly on small screen


StCyr:
**Describe the bug**

1. "posts, following, followers" line overlaps with the account's detail
2. followers and following accounts details don't wrap when screen's width is too small

See screenshot

**Proposed solution**

1. Put the "posts, following, followers" line below the account's detail
2. let followers and following accounts detail wrap to next line 

**Screenshots**

![image](https://user-images.githubusercontent.com/8179031/63422518-c8420e00-c40a-11e9-8541-acf45ba923c0.png)

**Client details:**
 - OS: iOS
 - Browser: FF
 - Device: iPhone7

PR #692, opened 2019-08-22
Better layout of the profile views on small screen

StCyr:
* Resolves: #690
* Target version: master 

### Summary

large screen:

image

small screen:

image


daita:
impressive work, thanks !

Issue #708, opened 2019-08-28
weird composer behaviour when hashtag or mention is last element in post


StCyr:
**Describe the bug**
When writing a post and put a hashtag at the end (just before clicking the "post..." button), the hashtag is broken.

**To Reproduce**
1. Go to Composer
2. write some text and end it with a hashtag or a mention
3. Directly after selecting the hastag/mention, post the toot
4. See that the hashtag is broken in the timeline

**Expected behavior**
The hashtag/mention shouldn't be broken

**Client details:**
 - OS: macOS
 - Browser: Firefox 68.0.2
 - Version: alpha 2.2
 - Device: Macbook Pro

**Possible cause**

I wonder if it's not linked to this issue: https://blog.francium.tech/vue-js-and-html5-contenteditable-attribute-issues-c727734c7684

PR #712, opened 2019-08-29
Fixes Composer's weird behavior when hashtag or mention is last element in post

StCyr:
* Resolves: #708
* Target version: master 

### Summary

vue-contenteditable updates its content when receiving the `oninput` event. 

Unfortunately, vue-tribute doesn't fire this event when it replaces a mention or hashtag. 

So, when we place a mention or hashtag using vue-tribute and directly click outside the Composer, the oninput event isn't fired and the post's content is reverted to what was typed before replacing the 
mention or hashtag.

This PR fixes this by firing the vue-contenteditable's oninput event upon reception of vue-tribute's `tribute-replaced`` event.


jancborchardt:
@daita @nextcloud/social up for review? :)

StCyr:
Here's a screencast of the issue:

Social-Nextcloud-MozillaFirefox2

Issue #157, opened 2018-12-07
Do not create Post if no field is filled


daita:
Check during post creation and cancel post creation if no field (to, cc, bcc) is filled 

example case: direct message to a non existant user.

jancborchardt:
> to, cc, bcc

:wink: this is Social … [→ Mail is over there](https://github.com/nextcloud/mail/).

PR #731, opened 2019-09-11
Prevent posting of direct message with no valid recipients

StCyr:
* Resolves: #157 
* Target version: master 

### Summary

See #157

### DIFF
--- a/lib/Db/StreamRequestBuilder.php
+++ b/lib/Db/StreamRequestBuilder.php
@@ -66,7 +66,7 @@ class StreamRequestBuilder extends CoreRequestBuilder {
 	 */
 	protected function getStreamInsertSql(): SocialQueryBuilder {
 		$qb = $this->getQueryBuilder();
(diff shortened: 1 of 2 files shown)


daita:
It is working slightly too much !

If I just type the Social Account of the recipient, without validating by pressing 'Enter' when the pop-in with the searched account shows up, I cannot post.

issue_recipients

StCyr:
> If I just type the Social Account of the recipient, without validating by pressing 'Enter' when the pop-in with the searched account shows up, I cannot post.

Ok, maybe I can validate automaticaly when there's only one account found and the user gets out of the tribute component without pressing 'Enter'....

StCyr:
> Ok, maybe I can validate automaticaly when there's only one account found and the user gets out of the tribute component without pressing 'Enter'....

Fixed: spacebar now works as the enter and tab keys.

StCyr:
should be good now

PR #744, opened 2019-09-16
Notification timeline improvements

violoncelloCH:
* Target version: master 

### Summary
polishing the notification timeline

- [x] use user component for "{account} followed you}"
- [x] add icons in front of to the action summary
- [ ] only show display name/username in action summary (and not complete handle)
- [ ] add more spacing between action summary and post or user component


jancborchardt:
@ArtificialOwl any comments on why this can’t be merged, or specific feedback for @violoncelloCH?

violoncelloCH:
it was not finished, but I think there were still some things in there (design) which would be worth it to port them over to a new PR if this one can't be merged

jancborchardt:
cc @artonge maybe worth to look at. @violoncelloCH it could be best to split the PR up in several smaller ones so it doesn’t get blocked? :)

Issue #773, opened 2019-10-01
search on #hastags are ignored


daita:
it seems that the front-end does not like search that starts with '#':

![Selection_110](https://user-images.githubusercontent.com/1038617/65959724-0c392180-e453-11e9-9ef3-69c4b95b7459.png)

As we can see, when typing an hashtag with the starting `#`, the request to the back-end contains empty search string.

PR #774, opened 2019-10-01
FIX: Removes heading '#' from the search terms as search won't work

StCyr:
FIX: Removes heading '#' from the search terms as search won't work when search term starts with such character (probably due to 'generateUrl')

Signed-off-by: Cyrille Bollu <cyrpub@bollu.be>

* Resolves: #773
* Target version: master


daita:
still an issue in the result age, the `@` or the `#` are not displayed:

Selection_112
Selection_113
Selection_114

Issue #781, opened 2019-10-03
In Composer, inserting an emoji inserts a line break before the emoji


StCyr:
**Describe the bug**
In Composer, inserting an emoji inserts a line break before the emoji

**Expected behavior**
The emoji should be inserted directly where the cursor is; no need for a line break

daita:
let's close this BR as we're trying to migrate emoji's component and we'll see if this issue is still there then

PR #791, opened 2019-10-06
Removes trailing '<br>' tag before inserting emoji (if any)

StCyr:
* Resolves: #781 
* Target version: master


daita:
not working as expected:

issue_emoji

Also, when adding an emoji, i need to type twice 'Enter' to jump a line

daita:
let's close this PR as we're trying to migrate emoji's component and we'll see if this issue is still there then

Issue #756, opened 2019-09-25
Single post - improvement


daita:
- [ ] Check viewer rights before returning data. @daita 
- [x] Remove the 'No posts found' message if there is no replies.
- [x] Display the original message with bigger font.
- [x] single post should only be triggered on a click on the 'date' and not the full post
- [x] Add post's author profile component
- [ ] Remove links to other nextcloud apps in header (when not authenticated).
- [ ] Error 401 when post has no replies
- [ ] Also show post's parents
- [ ] Error 500 when post in not public (eg: https://nextcloud.bollu.be/index.php/apps/social/@cyrille/1570354026755527405)
- [x] reply (and probably others) action don't work
- [x] Pictures not shown properly

daita:
Needs to check the list which bullet is already done, and which one is needed

PR #793, opened 2019-10-07
WIP: single post timeline improvement

StCyr:
* Resolves: #756 
* Target version: master 

### Summary

This is work in progress

### DIFF
--- a/src/mixins/accountMixins.js
+++ b/src/mixins/accountMixins.js
@@ -5,6 +5,8 @@
  *
  * @license GNU AGPL version 3 or any later version
  *
+ * @file provides global account related methods
+ *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as
  * published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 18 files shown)


jancborchardt:
@StCyr regarding the single post page, screenshots would be great for quick design review also from others. :)

src/mixins/accountMixins.js
```diff
@@ -0,0 +1,48 @@
...
+ * This file provides global account related methods
+ */
+
+import serverData from './serverData'
+
+export default {
+	mixins: [
+		serverData
+	],
+	methods: {
```
juliushaertl: Please use a computed property to make sure it is properly updated if needed

src/mixins/serverData.js
```diff
@@ -20,6 +20,22 @@
  *
  */
 
+/*
+ * This file provides global methods for using the serverData structure
+ * the serverData structure's data are initialy found in a DOM element
+ *  provided by the server.
+ *
+ * It may contain the following information:
```
juliushaertl: How about a jsdoc type definition for this? Then your IDE could give you hints when coding.

StCyr: ok. I'll do that

src/mixins/accountMixins.js
StCyr: ok

jancborchardt:
@StCyr what are the missing points here? :) Also there’s some conflicts to resolve

icewind1991:
rebased.

Haven't verified how much of the linked issue is closed with this.

before:
image

after:
image

there is still room for improvement but imo it makes sense to merge this as is and do further improvements in new PRs so we don't end up with stale PRs again

Issue #797, opened 2019-10-09
Page not found when trying to reply to a remote post


StCyr:
**Describe the bug
I wanted to reply to a post on a remote mastodon instance. The instance then prompted me for my account and redirected me to my social instance. But the redirection leads to a 404 error.

Missing interface?

![8D64A2CF-9D12-482C-BC4D-F4A074BD872D](https://user-images.githubusercontent.com/8179031/66492433-b13ba600-eab4-11e9-8390-106cae770bff.png)

daita:
you're low on battery !

SchBenedikt:
Automatically closed — replying to remote posts now works. `ActionService.php` processes reply actions and `NoteInterface` handles incoming Note activities properly.

PR #799, opened 2019-10-11
interact with remote items

ArtificialOwl:
This would allow to display the remote post locally when interaction is initiated from a remote instance: #797


karlitschek:
Closing after a triage of all open PRs: this 2019 branch predates several rewrites (`lib/webfinger.php` is gone, the timeline store and routing changed), so it can't be rebased in practice. Interacting with remote items from a remote-initiated flow is still a valid feature idea — tracked best as a fresh issue against the current code.

PR #827, opened 2019-11-22
Add named parameter for in_reply_to_prim to avoid NotNullConstraintViolationException

ljo:
Adding named parameter for in_reply_to_prim to avoid NotNullConstraintViolationException on InsertSql following the 20191001 migration updates in table.

* Target version: master


daita:
thanks for your input, i fixed it using the $qb->prim() on my side

PR #920, opened 2020-06-25
Make CI happy

juliushaertl:
As green ci is much nicer to have :wink:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,10 +2,10 @@ kind: pipeline
 name: compatibility
 steps:
   - name: compatibility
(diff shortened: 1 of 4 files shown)


juliushaertl:
>  we should probably also test on something newer than php 7.1 and NC 17, right? (but this can obviously be a separate PR) 

Ideally there would also be tests, because as of right now it only checks if the app can be enabled properly. :grin:

violoncelloCH:
should I create issues for this or we won't forget anyway?

daita:
you should create an issue

violoncelloCH:
here you go: #921 & #922

PR #982, opened 2020-08-22
Dark timeline entries for the dark Nextcloud theme

mashedkeyboard:
* Resolves: Issue not reported, just fixed
* Target version: master 

### Summary
The hover colour of timeline posts is always white at present, even with the dark theme enabled. This is very jarring for dark theme users. This PR contains an incredibly simple change to introduce a lighter dark hover highlight for dark theme users.

### DIFF
--- a/src/components/TimelineEntry.vue
+++ b/src/components/TimelineEntry.vue
@@ -138,6 +138,9 @@ export default {
 	.timeline-entry:hover {
 		background-color: #F5F5F5;
 	}
+	body.dark .timeline-entry:hover {
+		background-color: #202020;
+	}
 
 	.container-icon-boost {
 		display: inline-block;
--- a/src/components/TimelineEntry.vue
+++ b/src/components/TimelineEntry.vue
@@ -134,12 +134,9 @@ export default {
 	.timeline-entry {
 		padding: 10px;
 		margin-bottom: 10px;
-	}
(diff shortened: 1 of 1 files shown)


src/components/TimelineEntry.vue
```diff
@@ -138,6 +138,9 @@ export default {
 	.timeline-entry:hover {
 		background-color: #F5F5F5;
```
jancborchardt, line 139: If you simply use the variable `var(--color-background-hover)` here, they are both standardized and you don’t need the extra rule for the .dark class. :)

mashedkeyboard: That would, however, require me to know what I was doing :wink: sorted, should be good now! Thanks :smile:

jancborchardt (approved):
Looks good now! :) 

Something we should also do – but that’s something for a future pull request – is to make sure everything also works with keyboard-only. So have :focus states next to :hover, and make sure all the things you can click also work with enter. :)

jancborchardt:
Welcome to Nextcloud, @mashedkeyboard! We have a Nextcloud Talk instance with a channel for Social app development, where I could invite you with a guest account if you like? :)

mashedkeyboard:
Thank you, @jancborchardt! I'd like that.

PR #984, opened 2020-08-25
switch from vue-style-loader to style-loader

icewind1991:
vue-style-loader is unmaintained, broken with css-loader 4.x and regular style-loader now works properly with vue

this fixes styles not being loaded

PR #994, opened 2020-09-01
👌 IMPROVE: use vue Content components and transmit serverData over initialState

violoncelloCH:
- use vue components `Content` and `AppContent` with it's respective styling
  to be independent from server styling
- use OCP\IInitialStateService and @nextcloud/initial-state to transmit the serverData
  instead of using a hidden span element and jsonEncode/Decode
  this is needed to use the `Content` component, so the vue instance can be mounted to
  #content directly (otherwise server styling for #content interfers with the vue styling)
- also improves some general styling (mostly widths and margins/paddings)

~(please note that the styling isn't applied properly without #984 )~ -> rebased


lib/Controller/NavigationController.php
```diff
@@ -167,6 +170,7 @@ public function navigate(string $path = ''): TemplateResponse {
 					if ($cloudAddress !== null) {
 						$this->configService->setCloudUrl($cloudAddress);
 					} else {
+						$this->initialStateService->provideInitialState(Application::APP_NAME, 'serverData', $data['serverData']);
```
juliushaertl, line 173: Could still be simplified a bit since `$data` only has the `serverData` key, so it could just be a one-dimensional array in the code starting from line 147.

juliushaertl (approved):
Smaller comment, otherwise :+1:

lib/Controller/NavigationController.php
violoncelloCH: thanks! I've changed this as well...

juliushaertl:
:+1:

PR #314, opened 2019-01-07
Add the ability to follow from external website

daita:
related to #268

Implementation of OStatus (it seems to work like this):

- the remote website open a page (in popup) from the app, with the remote account in params.
- in the opened popup, we display the result of a search on the remote account set in params. We display also a 'Follow' button.
- if user is not authed, login/password.

ToDo:
- [x] Implement popup for following external accounts
- [x] Open popover to follow from public pages

## Follow from a public mastodon page
image

image

### Follow with a remote account from public profile pages:

image

PR #1027, opened 2020-09-14
render profile page for ostatus follow url

icewind1991:
when initiating a follow from a remote mastadon instance, the user gets directed to `apps/social/ostatus/follow/?uri=` which is currently broken.

this changes it to render the regular profile page for that link, allowing the user to follow the profile.

I'm not sure what the previous behaviour was before things got broken

### DIFF
--- a/lib/Controller/OStatusController.php
+++ b/lib/Controller/OStatusController.php
@@ -72,12 +72,15 @@ class OStatusController extends Controller {
 
 	/** @var IUserManager */
 	private $userSession;
+	/** @var IInitialStateService */
+	private $initialStateService;
 
 
 	/**
 	 * OStatusController constructor.
 	 *
 	 * @param IRequest $request
+	 * @param IInitialStateService $initialStateService
 	 * @param CacheActorService $cacheActorService
 	 * @param AccountService $accountService
 	 * @param CurlService $curlService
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):
Sure

lib/Controller/OStatusController.php
```diff
@@ -122,14 +125,14 @@ public function subscribe(string $uri): Response {
...
 				'currentUser' => [
-					'uid'         => $user->getUID(),
+					'uid' => $user->getUID(),
 					'displayName' => $user->getDisplayName(),
-				]
+				],
 			]);
 			return new TemplateResponse(
-				'social', 'main', 'guest'
+				'social', 'main', []
```
juliushaertl, line 135: Not sure when this changed but we had a separate entrypoint for this back then https://github.com/nextcloud/social/pull/314/files#diff-b5ae6be4c3786d95c10540fefacaac59
https://github.com/nextcloud/social/pull/314/files#diff-b5ae6be4c3786d95c10540fefacaac59

violoncelloCH: hmm I guess you mean what I've removed in #994 right?

jancborchardt:
@icewind1991 need to check your --signoff here as well for DCO to pass. :)

icewind1991:
done

daita (approved):

PR #1028, opened 2020-09-14
preserve some of the whitespace from html tags

icewind1991:
without the "html whitespace" a hashtag posted after a url will be seen as part of the link by linkify, and then mangled by the hashtagmangler and break the url.

additionally this keeps the appearance of the post closer to the original

Before:
image

After: 
image

Original:
image


daita (approved):

jancborchardt:
Good stuff @icewind1991 – can you check the --signoff so the DCO check passes?

icewind1991:
done

PR #1052, opened 2020-10-09
use default composer vendor directory

icewind1991:

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -46,6 +46,7 @@ use OCP\AppFramework\QueryException;
 use OCP\IServerContainer;
 use Throwable;
 
+require_once __DIR__ . '/../../vendor/autoload.php';
 
 /**
  * Class Application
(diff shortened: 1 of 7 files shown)


lib/AppInfo/Application.php
```diff
@@ -46,6 +46,7 @@
 use OCP\IServerContainer;
 use Throwable;
 
+require_once __DIR__ . '/../../vendor/autoload.php';
```
daita, line 49: Do we really want to move from the default composer/ if we need to add this line ?

daita (approved):

lib/AppInfo/Application.php
daita: I will merge this one, so we stop going back and forth with this. While waiting for @ChristophWurst point of view on the subject

ChristophWurst: This is fine IMO. I would not say that Nextcloud is responsible for registering a possible custom autoloader of the app, so this is the most appropriate location to do this.

PR #1068, opened 2020-10-12
show profile numbers inline instead of next to the name and avatar

icewind1991:
Before:
image
After:
image

I guess when a shorter name is used (I'll investigate why it isn't showing the display name) the `before` doesn't look as bad.

### DIFF
--- a/src/components/ProfileInfo.vue
+++ b/src/components/ProfileInfo.vue
@@ -22,23 +22,11 @@
 
 <template>
 	<div v-if="profileAccount && accountInfo" class="user-profile">
-		<div>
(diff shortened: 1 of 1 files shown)


jancborchardt:
Not sure about this, since it de-emphasizes that the Posts/Followers/following is actually a navigation?

The display name and the handle should for sure be directly below each other because they directly relate.

jancborchardt:
@icewind1991 when I refresh or force refresh on the profile page, I’m being redirected back to the Dashboard. Does that happen to you too? Makes it difficult to hard refresh :D

juliushaertl (approved):
Code makes sense, but didn't test

PR #1072, opened 2020-10-13
Search for nc20

daita:


daita:
Selection_018

first draft, we also need more feature in the UnifiedSearch so you can display result within the page itself, and action in the popup would be great

jancborchardt:
Can you add the avatar of the relevant person or post author to the search result? :)

PR #1136, opened 2020-11-22
Support unquoted fields in Signature header

mnd:

* Resolves: 
* Target version: master 

### Summary
When I try to subscribe to users from https://juick.com/ (e.g. @schors@juick.com) the server sends PUSH Accept message with the following Signature header:
```
keyId="https://juick.com/u/schors#main-key",created=1606061559,algorithm="rsa-sha256",headers="(request-target) host date digest",signature="<signature-data>"
``` 
where `created=1606061559` key-value pair has no quote around value part. Suggested change allows to part such signatures


daita:
Code looks good.

Issue #1133, opened 2020-11-22
Trying to follow creativecoding@libranet.de results in OCA\Social\Exceptions\ItemUnknownException


zeitgeist87:
Trying to follow `creativecoding@libranet.de` results in an `OCA\Social\Exceptions\ItemUnknownException`.  I added the following line just before the exception occurs to `/srv/http/cloud.mydomain.com/apps/social/lib/AP.php`

```
var_dump($this->get('type', $data, ''));
```

It prints out `string(5) "Group"`. Apparantly `Group` is an unknown type.

```
./occ social:account:follow -vvv zeitgeist creativecoding@libranet.de
string(5) "Group"

In AP.php line 383:
                                                
  [OCA\Social\Exceptions\ItemUnknownException]  
                                                

Exception trace:
  at /srv/http/cloud.mydomain.com/apps/social/lib/AP.php:383
 OCA\Social\AP->getItemFromType() at /srv/http/cloud.mydomain.com/apps/social/lib/AP.php:284
 OCA\Social\AP->getSimpleItemFromData() at /srv/http/cloud.mydomain.com/apps/social/lib/AP.php:216
 OCA\Social\AP->getItemFromData() at /srv/http/cloud.mydomain.com/apps/social/lib/Service/CurlService.php:207
 OCA\Social\Service\CurlService->retrieveAccount() at /srv/http/cloud.mydomain.com/apps/social/lib/Service/CacheActorService.php:218
 OCA\Social\Service\CacheActorService->getFromAccount() at /srv/http/cloud.mydomain.com/apps/social/lib/Service/FollowService.php:137
 OCA\Social\Service\FollowService->followAccount() at /srv/http/cloud.mydomain.com/apps/social/lib/Command/AccountFollowing.php:122
 OCA\Social\Command\AccountFollowing->execute() at /srv/http/cloud.mydomain.com/3rdparty/symfony/console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /srv/http/cloud.mydomain.com/core/Command/Base.php:169
 OC\Core\Command\Base->run() at /srv/http/cloud.mydomain.com/3rdparty/symfony/console/Application.php:1012
 Symfony\Component\Console\Application->doRunCommand() at /srv/http/cloud.mydomain.com/3rdparty/symfony/console/Application.php:272
 Symfony\Component\Console\Application->doRun() at /srv/http/cloud.mydomain.com/3rdparty/symfony/console/Application.php:148
[...]

PR #1138, opened 2020-11-23
Support all Actor types from Activity Vocabulary

mnd:
Support all types from https://www.w3.org/TR/2017/REC-activitystreams-vocabulary-20170523/

* Resolves: #1133 
* Target version: master 

### Summary
Support all actors from standard in same manner as for "Service" actor


daita:
Nice. Thanks for your work !

PR #1090, opened 2020-10-19
Mastadon api fixes

icewind1991:
Various api fixes needed to allow the frontend to use this api

PR #1148, opened 2020-11-28
Store newlines as <br /> tag in new posted Notes

### LINKED ISSUE #1139: Improve support for multiline notes
**Is your feature request related to a problem? Please describe.**
When I writing new Note with embedded newlines they displayed on web interface as single line message. This happens because newlines stored in DB and returned to HTML code as is without replacement to `</br>` tag

**Describe the solution you'd like**
**Describe alternatives you've considered**
I'm not expert in current fediverse infrastructure and can't decide which solution would be better:
1. Replace newlines with `</br>` when post created and stored to DB. In this case all recipients will got message with HTML markup.
2. Replace newlines with `</br>` when post displayed on WEB interface. In this case recipients will got message as it was written, but all newlines will be displayed correctly on the web.

mnd:
ActivityStream assumes that content is HTML message: https://www.w3.org/TR/2017/REC-activitystreams-vocabulary-20170523/#dfn-content
To fullfill this assumption save newlines as `<br />` during post creation

* Resolves: #1139  
* Target version: master 

### Summary
Suggested change should allow to store post in manner that allows to display it correctly on the web. Looks like other fediverse instances behave in similar manner. My only doubt is what better: use `nl2br` php function to have both newlines and `<br \>` tag, or use `replace()` call to have only `<br \>` tag

### DIFF
--- a/lib/Model/ActivityPub/ACore.php
+++ b/lib/Model/ActivityPub/ACore.php
@@ -566,7 +566,7 @@ class ACore extends Item implements JsonSerializable {
 				return $value;
 
 			case self::AS_CONTENT:
-				$value = strip_tags($value, ['a', 'p', 'span', 'br']);
+				$value = strip_tags($value, '<a><p><span><br>');
 				$value = html_entity_decode($value, ENT_QUOTES | ENT_HTML5);
 
 				return $value;
--- a/lib/Service/PostService.php
+++ b/lib/Service/PostService.php
@@ -126,7 +126,7 @@ class PostService {
 		$this->streamService->assignItem($note, $actor, $post->getType());
 
 		$note->setAttributedTo($actor->getId());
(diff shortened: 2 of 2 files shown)


mnd:
Updated according to improvements from #1090

mnd:
> My only doubt is what better: use nl2br php function to have both newlines and <br \> tag, or use replace() call to have only <br \> tag

Replace is better because mastodon uses "white-space: pre-wrap" as default style and using `<br \>\n` will look ugly on mastodon web.

karlitschek:
Closing after a triage of all open PRs: the Mastodon API path now converts newlines (`nl2br` in `ApiController`, pinned by a unit test), and the composer component this patched has been rewritten. The one remaining gap — the app's own local post path in `PostService` — will be fixed in a fresh follow-up. Thanks for the patch.

PR #1179, opened 2021-01-10
Allow to follow to PeerTube 3.0.0 users

mnd:
Solved three issues:
* lib/Service/ConfigService.php: With "Accept: application/json" PeerTube
  returns user page in HTML format. Replace header with
  "Accept: application/ld+json".
* lib/Service/CurlService.php: Do not broke "$host" variable if host-meta
  resultion failed.
* lib/Service/SignatureService.php: POST to https://framatube.org/accounts/fsf/inbox
  ends with error "(request-target) was not a signed header". Added
  (request-target) to signature.

* Resolves: #1133 
* Target version: master 

### Summary
I tried to communicate with updated PeerTube instance that contains bugfix mentioned in #1133 and observed additional issues. After suggested changes I was able to follow PeerTube account

### DIFF
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -444,7 +444,7 @@ class ConfigService {
 
 		if ($request->getType() === Request::TYPE_GET) {
 			$request->addHeader(
-				'Accept', 'application/json; profile="https://www.w3.org/ns/activitystreams"'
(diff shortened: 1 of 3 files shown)


CarlSchwan (approved):
sorry for the delay

PR #1180, opened 2021-01-10
Patch CI (and Cypress)

elgorro:
* Resolves: # Fixing Pipeline
* Target version: master 

### Summary
- Version Cypress-Workflow = 2
- Support NodeJS 14
- Update Cypress packages to latest with dependencies


elgorro:
The pipeline is broken due Github-Workflow changes. Would appreciate help with "cypress". Found this topic: https://stackoverflow.com/questions/55501838/how-exactly-do-before-and-beforeeach-work-in-cypress.

I'm totally new to cypress:
1. Updated the Github-Task-Version from cypress
2. Upgraded NPM package (and some other for testing)
3. Run into trouble with cypress - I'm not sure if this is the correct behavior or not.

elgorro:
Something is strange. Here as well: https://stackoverflow.com/questions/65652108/cypress-uncaught-typeerror-cannot-read-property-of-undefined-json-file

artonge:
Hello @elgorro, this should be fixed in the current master. Can you check and close if it is resolved ?

PR #1183, opened 2021-01-13
Display Boosted messages in user's timelime

### LINKED ISSUE #697: Repost does not appear on my wall
**Describe the bug**
Repost does not appear on my wall

**To Reproduce**
1. I click on repost
2. Request leaves
3. Response comes, status 200
4. The button lights up in blue
5. Repost does not appear on my wall

**Expected behavior**
Repost should appear on my profile

**Client details:**
 - OS: Ubuntu
 - Browser Firefox
 - Version 68.0.2
 - Device: desktop

mnd:

* Resolves: #697
* Target version: master 

### Summary
This is non-ideal solution for the issue, but this solution solves the issue. Drawbacks of the solution:
1. Looks like it should generic replacement function for limitToAttributedTo instead of suggested `andWhere` clause
2. Probably Boost's should be counted in the `countNotesFromActorId` function as well.
3. Probably `'Announce'` string should be replaced with `Announce::TYPE` constant

### DIFF
--- a/lib/Db/SocialCrossQueryBuilder.php
+++ b/lib/Db/SocialCrossQueryBuilder.php
@@ -255,8 +255,8 @@ class SocialCrossQueryBuilder extends SocialCoreQueryBuilder {
 	/**
 	 * @param string $alias
 	 */
-	public function leftJoinStreamAction(string $alias = 'sa'): void {
-		if ($this->getType() !== QueryBuilder::SELECT || !$this->hasViewer()) {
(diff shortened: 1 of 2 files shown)


karlitschek:
Closing after a triage of all open PRs: boosts now appear in account timelines on master — `Announce` items are stored with `attributed_to` set to the boosting actor and the account timeline was rewritten on the ProbeOptions query path, so the code this patches no longer exists. Thanks for the contribution.

PR #1230, opened 2021-03-24
Do not use wrong content type for GET requests

vitalyster:
See https://www.w3.org/TR/activitypub/#retrieving-objects


karlitschek:
Closing after a triage of all open PRs: master now sends the proper `Accept` header on GET and `Content-Type` on POST in `ConfigService::configureRequest()`, so the wrong-content-type issue this addressed is fixed. Thanks for the report and the patch.

Issue #1203, opened 2021-02-07
could not extract app social


wesleyehowell:
Trying to install the app into Ubuntu 20.04 Nextcloud snap, 20.0.7snap1 and I'm told it couldn't extract the app.

I'm using Ubuntu 20.04, with Firefox 85.

Nextcloud log:

Error | settings | Exception: Could not extract app social/snap/nextcloud/26119/htdocs/apps/settings/lib/Controller/AppSettingsController.php - line 445: OC\Installer->downloadApp("social")/snap/nextcloud/26119/htdocs/lib/private/AppFramework/Http/Dispatcher.php - line 169: OCA\Settings\Controller\AppSettingsController->enableApps([ "social"], [])/snap/nextcloud/26119/htdocs/lib/private/AppFramework/Http/Dispatcher.php - line 100: OC\AppFramework\Http\Dispatcher->executeController(OCA\Settings ... {}, "enableApps")/snap/nextcloud/26119/htdocs/lib/private/AppFramework/App.php - line 152: OC\AppFramework\Http\Dispatcher->dispatch(OCA\Settings ... {}, "enableApps")/snap/nextcloud/26119/htdocs/lib/private/Route/Router.php - line 309: OC\AppFramework\App::main("OCA\\Settin ... r", "enableApps", OC\AppFramew ... {}, { _route: "s ... "})/snap/nextcloud/26119/htdocs/lib/base.php - line 1008: OC\Route\Router->match("/settings/apps/enable")/snap/nextcloud/26119/htdocs/index.php - line 37: OC::handleRequest()
-- | -- | --

End of section

jaitaiwan:
I'm also seeing the same thing

Aarontheissueguy:
Same issue here

senooken:
Hi. I had same error too on Nextcloud 20.0.7 git. I avoided this error by following manual install.

```
cd apps
wget https://github.com/nextcloud/social/releases/download/v0.4.2/social-0.4.2.tar.gz
tar -xf apps/social-0.4.2.tar.gz
cd ..
php occ app:enable social
```

Aarontheissueguy:
How would one do this if he uses the snap.

Ascariota:
I have also this issue 

`Exception: Could not extract app social
/var/www/nextcloud/apps/settings/lib/Controller/AppSettingsController.php - line 445:

OC\Installer->downloadApp()

[...]

Issue #1016, opened 2020-09-11
using WebfingerManager


ArtificialOwl:
- fix #121 
- nextcloud/server#21817

WIP as it might be better to create a WellKnownManager that would manage 
- /.well-known/webfinger
- /.well-known/nodeinfo
- /.well-known/hostmeta

instead of just managing the first one

daita:
Uncomment https://github.com/nextcloud/social/blob/master/lib/AppInfo/Application.php#L76 to enable the feature, once nextcloud/server#21817 is merged

tacruc:
On nextcloud 21 I recive the .well-known/webfinger not correct.


```
curl https://domain.com/.well-known/webfinger?resource=acct:user@domain.com
{"message":"webfinger not supported"}
```

I was wounding if it is connected to this Issue here?

daita:
not related, this will be fixed later this week !

mashedkeyboard:
> On nextcloud 21 I recive the .well-known/webfinger not correct warning on the social page.
> 
> ```
> curl https://domain.com/.well-known/webfinger?resource=acct:user@domain.com
> {"message":"webfinger not supported"}
> ```
> 
> I was wounding if it is connected to this Issue here?

I'm still geting this issue on NC21 - @daita, do you know when that fix is planned for inclusion in a release?

cClaude:
Getting ```{"message":"webfinger not supported"}``` as well on stable channel.

This message is not clear for me. I would like to know what is wrong in my configuration ?

mashedkeyboard:
social seems to be completely incompatible with Nextcloud 21 until this is fixed, which is a shame :( any word on a fix, @ArtificialOwl? I can work on a PR if it would be helpful

CarlSchwan:
This is done

PR #1283, opened 2021-06-24
Update version on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


mashedkeyboard:
This should not be done until the app is actually fixed. The upgrade path is broken https://github.com/nextcloud/social/issues/1203, and even with a fresh install the app doesn't actually work because the webfinger setup changed in NC21 https://github.com/nextcloud/social/issues/1016.

PR #1369, opened 2022-04-14
Make social app compatible with 24

CarlSchwan:
Due to the difference between 23 and 24 in regard to the query builder private API it is a bit tricky to support both 24 and 23 at the same time

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -35,7 +35,7 @@
 		<database>pgsql</database>
 		<database>sqlite</database>
 		<database>mysql</database>
-		<nextcloud min-version="20" max-version="21"/>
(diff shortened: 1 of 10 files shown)


skjnldsv:
It means thre is no version for 22 and 23

ArtificialOwl (approved):
looks good

mirenbz:
So does this mean it is supposed to work with 24?

McSinyx:
Could a maintainer please tag a release?

ArtificialOwl:
This mean a version for NC24 will be released; just not yet

MagicTrevor:
Looking forward to it.

PR #1371, opened 2022-04-14
Update version on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


CarlSchwan:
I updated to support both 24 and 25 in https://github.com/nextcloud/social/pull/1377 (this makes the development easier even though master targets 24)

PR #1393, opened 2022-04-20
Fix JrdResponse not containing template

CarlSchwan:
For now this is modified JrdResponse from
https://github.com/nextcloud/server/pull/26178

Let's hope that we can switch to the OCP one in NC 25

### DIFF
--- a/lib/WellKnown/JrdResponse.php
+++ b/lib/WellKnown/JrdResponse.php
@@ -1,0 +1,174 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2020 Christoph Wurst <christoph@winzerhof-wurst.at>
+ *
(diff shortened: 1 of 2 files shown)


lib/WellKnown/JrdResponse.php
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * A JSON Document Format (JDF) response to a well-known request
+ *
+ * @ref https://tools.ietf.org/html/rfc6415#appendix-A
+ * @ref https://tools.ietf.org/html/rfc7033#section-4.4
+ */
+final class JrdResponse implements IResponse {
+	private string $subject;
+	private ?String $expires = null;
```
ArtificialOwl, line 42: upper case `String` ?

ArtificialOwl (approved):
not tested, but code looks good.

PR #1416, opened 2022-05-11
Create instance if it doesn't exist

CarlSchwan:


lib/Db/InstancesRequest.php
```diff
@@ -54,7 +55,19 @@ public function save(Instance $instance) {
...
+			->setValue('title', $qb->createNamedParameter($instance->getTitle()))
+			->setValue('version', $qb->createNamedParameter($instance->getVersion()))
+			->setValue('short_description', $qb->createNamedParameter($instance->getShortDescription()))
+			->setValue('description', $qb->createNamedParameter($instance->getDescription()))
+			->setValue('email', $qb->createNamedParameter($instance->getEmail()))
+			->setValue('urls', $qb->createNamedParameter(json_encode($instance->getUrls())))
+			->setValue('stats', $qb->createNamedParameter(json_encode($instance->getStats())))
+			->setValue('usage', $qb->createNamedParameter(json_encode($instance->getUsage())))
+			->setValue('image', $qb->createNamedParameter($instance->getImage()))
+			->setValue('languages', $qb->createNamedParameter(json_encode($instance->getImage())))
```
CarlSchwan, line 69:
```suggestion
			->setValue('languages', $qb->createNamedParameter(json_encode($instance->getLanguages())))
```

lib/Service/InstanceService.php
```diff
@@ -36,33 +36,40 @@
...
-	public function createLocal(): void {
+	public function createLocal(): Instance {
+		$instance = new Instance();
+		$instance->setLocal(true)
+			->setVersion($this->config->getAppValue('social', 'installed_version', '0.0'))
+			->setApprovalRequired(false)
+			->setDescription($this->config->getAppValue('theming', 'slogan', 'a safe home for your data'))
+			->setTitle($this->config->getAppValue('theming', 'name', 'Nextcloud Social'));
+		$this->instancesRequest->save($instance);
+		return $instance;
```
ArtificialOwl, line 69: we need to implement on the front-end a way to assign a contactAccount to the `Instance`

CarlSchwan: And also to add all the other fields. Could you create an issue so that we don't forget?

ArtificialOwl (approved):

lib/Service/InstanceService.php
ArtificialOwl: done

PR #1455, opened 2022-08-07
Update npm and node engines versions

nextcloud-command:
Automated update of the npm and node engines versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,125 +1,125 @@
 {
-	"name": "social",
-	"description": "🎉 Nextcloud becomes part of the federated social networks!",
-	"version": "0.0.1",
-	"authors": [
-		{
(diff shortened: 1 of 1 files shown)


CarlSchwan (approved):

max-nextcloud (changes requested):
I see two issues with this:

* skjnldsv/read-package-engines-version-actions does not seem to like the version expression for npm (see  https://github.com/nextcloud/social/runs/7740827926?check_suite_focus=true )
* node-sass needs to be at least version 6 to support node 16. In order to use node-sass 6, we need sass-loader >= 11, which in turn requires webpack >= 5.

CarlSchwan:
Done in a seperate PR

PR #1456, opened 2022-08-09
Fix/ci

max-nextcloud:
checking if the updated node and npm versions actually fix the build


CarlSchwan:
Fixed when moving to nc vue component 7

PR #1506, opened 2022-11-02
returns 404 on missing Actor

ArtificialOwl:
would be even better to send empty response, instead of empty string.

### DIFF
--- a/lib/WellKnown/JrdResponse.php
+++ b/lib/WellKnown/JrdResponse.php
@@ -26,6 +26,8 @@ declare(strict_types=1);
 
 namespace OCA\Social\WellKnown;
 
+use OCP\AppFramework\Http;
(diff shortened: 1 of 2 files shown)


lib/WellKnown/JrdResponse.php
```diff
@@ -151,13 +164,17 @@ public function addLink(string $rel,
...
 	 */
 	public function toHttpResponse(): Response {
-		return new JSONResponse(array_filter([
-			'subject' => $this->subject,
-			'expires' => $this->expires,
-			'aliases' => $this->aliases,
-			'properties' => $this->properties,
-			'links' => $this->links,
-		]));
+	$data =	array_filter(
```
CarlSchwan, line 167: Identation 😜

CarlSchwan (approved):

PR #1572, opened 2023-01-12
Improve rendering of user entry in user lists

artonge:


artonge:
Can't reproduce cypress issue locally, and it is not due to this PR. Merging.

ArtificialOwl (approved):

PR #1575, opened 2023-01-12
search on uri

ArtificialOwl:


lib/Service/CacheActorService.php
```diff
@@ -152,7 +152,7 @@ public function getFromId(string $id, bool $refresh = false): Person {
 				throw new InvalidResourceException();
 			}
 
-			if ($id !== $actor->getId()) {
+			if (parse_url($id, PHP_URL_HOST) !== parse_url($actor->getId(), PHP_URL_HOST)) {
```
artonge, line 155: From what I understand, this switch from comparing `username@hostname` to comparing `hostname`.

Am I right ?

ArtificialOwl: we were comparing the full url before but because we are doing the getFromId() on an alias, it is not working anymore. This was added to avoid a random instance to returns an account from another instance. so we just need to check on the domain name that alias belongs to same instance

artonge (approved):

PR #1585, opened 2023-01-17
local acct is username

ArtificialOwl:


lib/Model/ActivityPub/Actor/Person.php
```diff
@@ -721,7 +721,7 @@ public function exportAsLocal(): array {
 		$result =
 			[
 				"username" => $this->getPreferredUsername(),
-				"acct" => $this->getAccount(),
+				"acct" => $this->isLocal() ? $this->getPreferredUsername() : $this->getAccount(),
```
artonge, line 724: Isn't `preferredUsername` the display name and not the username ?

artonge: Just checked, and I think you are right !

artonge (approved):

PR #1588, opened 2023-01-19
Fix posting attachments

artonge:
- Prevent posting more than one attachment
- Fix rendering of attachments
- Update frontend and backend to properly send and accept array of post data for mentions, hashtags, and attachments


PVince81 (approved):
:+1: code looks fine

would be good if you could post screenshots in your PRs

PR #1616, opened 2023-02-28
Use Pleroma-compatible version string

mkljczk:
### Summary

As documented here: https://gitlab.com/soapbox-pub/soapbox/-/blob/develop/docs/development/developing-backend.md

This is implemented by other non-Mastodon backends utilizing Mastodon API, including Pixelfed, Pleroma, Mitra, Friendica…


lib/Model/Instance.php
```diff
@@ -92,7 +92,7 @@ public function setTitle(string $title): self {
 	}
 
 	public function getVersion(): string {
-		return $this->version;
+		return '4.1.0 (compatible; NextCloud ' . $this->version . ')';
```
ArtificialOwl, line 95: please make it 'Nextcloud Social' instead of 'NextCloud'

artonge (approved):

PR #1646, opened 2023-03-12
fix(l10n): Fixed grammar

rakekniven:
Reported at Transifex.


artonge (approved):
:see_no_evil: thanks !!

Issue #1701, opened 2023-03-24
Unable to open Profile


srijansaxena11:
**Describe the bug**
When trying to open profile it gets stuck at loading the page.

**To Reproduce**
1. Go to social app
2. Click on Profile

**Expected behavior**
Profile should open.

But it doesn't and the page keeps on loading.

**Screenshots**
![image](https://user-images.githubusercontent.com/34964694/227533903-21524988-4b1f-44ff-9502-b7b86fc7ba88.png)

**Client details:**
 - OS: Windows
 - Browser: Microsoft Edge
 - Version 11
 - Device: Desktop

<details>
<summary>Server details</summary>

**Social app version:** 0.6.0-beta5

**Operating system**: Ubuntu 22.04.1 LTS

**Web server:** Apache

**Database:** MySQL

**PHP version:** 8.1.2

**Nextcloud version:** (see Nextcloud admin page) 26.0.0
</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

vue.runtime.esm.js:3049 
        
       TypeError: Cannot read properties of undefined (reading 'id')
    at Profile.vue:100:1
    at p (Profile.vue:44:1)
    at Generator.<anonymous> (Profile.vue:44:1)
    at Generator.next (Profile.vue:44:1)
    at p (Profile.vue:44:1)
    at a (Profile.vue:44:1)
```

</details>

roelfrenkema:
Same here also webfinger seems not to work. Related?

srijansaxena11:
> Same here also webfinger seems not to work. Related?

Webfinger works fine. The profile still doesn't work though.

srijansaxena11:
Do we have any update on this? Anything to try out? Any updates or anything under development to fix it?

ArtificialOwl:
can you replicate the issue after installing https://github.com/nextcloud-releases/social/releases/download/v0.6.0-rc1/social-v0.6.0-rc1.tar.gz ?

srijansaxena11:
> can you replicate the issue after installing https://github.com/nextcloud-releases/social/releases/download/v0.6.0-rc1/social-v0.6.0-rc1.tar.gz ?

Hi @ArtificialOwl . I am on rc1 only: 
[...]

PR #1732, opened 2023-04-12
Update Profile.vue to read this.id instead of this.accountInfo.id

srijansaxena11:
* Resolves: #1701 
* Target version: master 

### Summary
As explained in the mentioned issue, on opening the Profile, Profile.vue is trying to access this.accountInfo.id which is not found in the response returned by the request. The response JSON has id which technically should be accessible directly via this.id. Hence the change from this.accountInfo.id to this.id.

### DIFF
--- a/src/views/Profile.vue
+++ b/src/views/Profile.vue
@@ -97,7 +97,7 @@ export default {
 		/** @type {[import('../types/Mastodon').Account]} */
 		const response = await this.$store.dispatch(fetchMethod, this.profileAccount)
 		this.uid = response.acct
-		await this.$store.dispatch('fetchAccountRelationshipInfo', [this.accountInfo.id])
+		await this.$store.dispatch('fetchAccountRelationshipInfo', [this.id])
 	},
 }
 </script>


artonge:
Hey @srijansaxena11, thanks for the PR.

Without diving into the code, it is a hard to understand what the change provides. Can you add a description to your PR and set a appropriate title?

srijansaxena11:
> Hey @srijansaxena11, thanks for the PR.
> 
> Without diving into the code, it is a hard to understand what the change provides. Can you add a description to your PR and set an appropriate title?

I have updated it. Please review the change as well.

artonge:
Thanks for the extra info :).

I think something might be wrong somewhere else, as the result of `https://nextcloud_server1.test/apps/social/api/v1/global/account/info?account=admin@nextcloud_server1.test` should contain an `id`.

Can you post the content of your request?

You should have something similar to:

```json
{
	"id": "1",
	"username": "admin",
	"acct": "admin",
	"display_name": "admin",
	"locked": false,
	"bot": false,
	"discoverable": false,
	"group": false,
	"created_at": "2023-04-13T15:37:58.000Z",
	"note": "",
	"url": "https:\/\/nextcloud_server1.test\/apps\/social\/@admin",
	"avatar": "https:\/\/nextcloud_server1.test\/apps\/social\/media\/56102672-6120-44c0-ad4c-be81ca0a261b",
	"avatar_static": "https:\/\/nextcloud_server1.test\/apps\/social\/media\/56102672-6120-44c0-ad4c-be81ca0a261b",
	"header": "https:\/\/nextcloud_server1.test\/apps\/social\/media\/56102672-6120-44c0-ad4c-be81ca0a261b",
	"header_static": "https:\/\/nextcloud_server1.test\/apps\/social\/media\/56102672-6120-44c0-ad4c-be81ca0a261b",
	"followers_count": 0,
	"following_count": 0,
	"statuses_count": 0,
	"last_status_at": "",
	"source": {
		"privacy": "public",
		"sensitive": false,
		"language": "en",
		"note": "",
		"fields": [],
		"follow_requests_count": 0
	},
	"emojis": [],
	"fields": []
}
```

artonge:
`this` refers to the Vue component, not the response.

srijansaxena11:
Will this somehow resolve the issue?:

```
if (response && response.acct){
            this.uid = response.acct
            if (this.accountInfo && this.accountInfo.id) {
                        await this.$store.dispatch('fetchAccountRelationshipInfo', [this.accountInfo.id])
            }
}
```

artonge:
I don't think so. It will make the error disappear, but it won't fix the loading of the profile.
Do you have any other error?

srijansaxena11:
> I don't think so. It will make the error disappear, but it won't fix the loading of the profile. Do you have any other error?

No. Just this one only for now. Once I can follow some accounts then only, I can test further. But for now, I am stuck here only. I am also trying to setup my local setup to try to figure out the issue but webfinger is not working on localhost for some reasons.

PR #1735, opened 2023-04-14
Retrieving id from response instead of this.accountInfo

srijansaxena11:
As this.accountInfo is coming up as undefined, but the response does contain the id of the account, it should work fine in passing the id from response instead of this.accountInfo. this.accountInfo is undefined for localaccount because it is trying to find username@nextcloud.domain user from state which has username@localhost user.

* Resolves: #1701 
* Target version: master


srijansaxena11:
No longer required after #1738 is merged.

srijansaxena11:
No longer required after #1738 is merged.

PR #1738, opened 2023-04-15
Update account.js

srijansaxena11:
The method _getActorIdForAccount was not returning the value. Return statement added. Also the state.accountIdMap contains local accounts as username@localhost. But the account passed to it was username@nextcloud.domain even in case of local accounts. The same has been handled by checking the hostname of account to the current hostname. Should work fine for local account.

* Resolves: #1701 
* Target version: master

### DIFF
--- a/src/store/account.js
+++ b/src/store/account.js
@@ -53,7 +53,7 @@ const addAccount = (state, { actorId, data }) => {
 	const accountId = (data.acct.indexOf('@') === -1) ? data.acct + '@' + new URL(data.url).hostname : data.acct
 	set(state.accountIdMap, accountId, data.url)
 }
-const _getActorIdForAccount = (account) => state.accountIdMap[account]
+const _getActorIdForAccount = (account) => {return state.accountIdMap[account]}
 
 /** @type {import('vuex').MutationTree<state, any>} */
 const mutations = {
@@ -134,6 +134,11 @@ const getters = {
 	},
 	getAccount(state, getters) {
 		return (/** @type {string} */ account) => {
+			var currentNextcloudDomain = window.location.hostname
+			var accountArray = account.split("@")
+			var accountName = accountArray.at(0)
+			var accountDomain = accountArray.at(-1)
+			account = (accountDomain == currentNextcloudDomain) ? accountName + "@localhost" : account
 			return state.accounts[_getActorIdForAccount(account)]
 		}
 	},


srijansaxena11:
Can anyone please review and merge this PR and discuss about other issues?
@ArtificialOwl @artonge

ArtificialOwl:
@artonge ?

artonge:
Hello @srijansaxena11, I think the issue is more with your config than with the code.

Can you check the value of `overwrite.cli.url` in your `config.php`?
It should be the same as the domain you are using to use your Nextcloud server.

artonge:
You will probably have to use the `occ social:reset` change the domain name the Social is using.

srijansaxena11:
> Hello @srijansaxena11, I think the issue is more with your config than with the code.
> 
> Can you check the value of `overwrite.cli.url` in your `config.php`? It should be the same as the domain you are using to use your Nextcloud server.

overwrite.cli.url is my domain name only. 



> You will probably have to use the `occ social:reset` change the domain name the Social is using.

occ social:check:install also shows the cloud URL as my domain name only. I already did the social:reset as my first step to debug the issue.

srijansaxena11:
Do I need to check somewhere in any of the tables in database?

artonge:
I do not have a publicly available setup.

Just to be sure that I understand the issue, `state.accountIdMap` contains an entry for `localUserId@localhost`, which should be `localUserId@domain.tld`.

The front end [is using the host part of the URL to fill the missing host part for local users](https://github.com/nextcloud/social/blob/d2e137133f08d53695a7376625dff0b2ee64139d/src/store/account.js#L53).

But this is for local accounts, what is the issue with remote ones?

srijansaxena11:
> Just to be sure that I understand the issue, `state.accountIdMap` contains an entry for `localUserId@localhost`, which should be `localUserId@domain.tld`.

Correct. Or vice-versa.


> But this is for local accounts, what is the issue with remote ones?

The very first description of this thread which I wrote. The _getActorIdForAccount method is not returning any value. Adding the return statement to it fixes the issue with remote accounts.

artonge:
> Correct. Or vice-versa.

What do you mean?

> The very first description of this thread which I wrote. The _getActorIdForAccount method is not returning any value. Adding the return statement to it fixes the issue with remote accounts.

The original code does return a value, there is no need for a `return` statement for arrow functions.

srijansaxena11:
> What do you mean?

like the account to be searched and the account in accountIdMap should both have either @domain.com or @localhost.



> The original code does return a value, there is no need for a `return` statement for arrow functions.

Not sure about that but adding the return statement resolves it and removing it brings it back. I set up my local nextcloud social setup to verify these small things.

srijansaxena11:
The `accountInfo()` from `accountMixins.js` points to undefined value without the return statement in `_getActorIdForAccount` in `account.js`

ArtificialOwl:
> 



> @artonge Do you have it working somewhere where I can join and test? I will be more than happy to identify the issue. Also the follow feature is not working. I am also trying to identify the issue there.

Send me a mail to maxence@nextcloud.com i'll generate an account on our testing installation.
Can you also send me your social account also so I can check your setup from outside ?

srijansaxena11:
> Send me a mail to [maxence@nextcloud.com](mailto:maxence@nextcloud.com) i'll generate an account on our testing installation.

Sure. What am I supposed to mention?


> Can you also send me your social account also so I can check your setup from outside ?

My social is @srijan@cloud.srijan.site. See if it is found from outside. Another issue is I am unable to follow any user (local or global). It might be related to webfinger as someone else pointed out in another issue thread. But doing the suggested changes didn't resolve that. The changes were the response of webfinger was missing `acct:` in subject whereas the response of mastadon and friendica had it in subject. But that didn't help too.

srijansaxena11:
> Send me a mail to [maxence@nextcloud.com](mailto:maxence@nextcloud.com) i'll generate an account on our testing installation.

I have dropped the mail.

srijansaxena11:
Hi @ArtificialOwl @artonge 
I made progress. I can be followed now from other accounts (Social, Mastodon, etc.) but I cannot follow any account. I get the following error:
image

Any suggestions to debug this?

PR #1748, opened 2023-04-25
Modernize e2e tests

artonge:


ArtificialOwl:
commit that fix:

- retrieve GET params using the URL instead of `getParams()`
- do not install `symfony/console` to avoid conflict with core (composer)
- enforce table alias (sqlite)

PR #1775, opened 2023-06-09
fill status.actor.avatar

ArtificialOwl:
- fix empty status.account.avatar on timeline notification on like
- fix empty status.account.avatar on timeline notification on mention
- fix empty status.account.avatar on timeline notification on boost

@artonge however, it does not show anything on the front-end 

Selection_074
Selection_075


artonge:
Tried to reproduce, but could even get the avatar in the response.

artonge (approved):

PR #1785, opened 2023-06-14
disable https for tests

ArtificialOwl:
fully disable https when running tests


artonge:
Not needed in the end

PR #1951, opened 2024-11-21
fix: Handle when required `resource` parameter is missing or empty

### LINKED ISSUE #816: HTTP status on .well-known/webfinger etc
0.2.101

* If invoked with no resource parameter at all, the response is 200 with content type HTML, and a plain-text error message. Example: `curl http://host/.well-known/webfinger`

This is very confusing. I don't know what the webfinger spec says, but I would suggest either 404, or 400.

Secondly: An error message is fine, but HTML?

Compare:

* If invoked with a non-existing resource, the response is 404. Example: `curl -v http://host/.well-known/webfinger?resource=does-not-exist`

Also: The self-check [here](https://github.com/nextcloud/social/blob/bad0024e06ec63f21ef4323b68ae9d49fe167e68/lib/Service/CheckService.php#L317) checks for 200 on the URL without any resource argument. This seems not such a good plan because:

* the request you are testing is not used for anything useful, so whether or not it works is not relevant;
* you don't test whether a valid JSON file actually comes back for an existing account;
* and that's particularly important because your instructions use `.htaccess` hacking to redirect the original request to a PHP script with an extra URL argument -- lots can go wrong when this is mapped, in particular because the `?resource` in the original request now needs to be a `&resource` as it is appended to some other argument.

### LINKED ISSUE #1888: Webfilnger giving Internal Server Error and exception error in logs
**Describe the bug**
Going to the url for webfinger on my Nextcloud gives an internal server error. And when I check the logs, I see the following message:

Exception parse_str(): Argument #1 ($string) must be of type string, null given in file '/opt/nextcloud/apps/social/lib/WellKnown/WebfingerHandler.php' line 235

**To Reproduce**
1. Go to https://cloud.home.dragongames.net/index.php/.well-known/webfinger
2. See error

**Expected behavior**
I'm not entirely sure what is supposed to happen as I've never been able to get this to work. But I do know that the 500 error and the error in the log file are not correct

**Screenshots**
n/a my site is public ish

**Client details:**
 - OS: Fedora 39
 - Browser: Chrome
 - Version: 29
 - Device: Laptop

<details>
<summary>Server details</summary>

**Social app version:**
8.0
**Operating system**:
Debian 12.5
**Web server:**
Apache2
**Database:**
MariaDB
**PHP version:**
8.2
**Nextcloud version:** (see Nextcloud admin page)
29
</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Exception parse_str(): Argument #1 ($string) must be of type string, null given in file '/opt/nextcloud/apps/social/lib/WellKnown/WebfingerHandler.php' line 235

#### Browser log
```
n/a this is all server side

</details>

joshtrichards:
* Resolves: #1888 
* Target version: master 

### Summary

`resource` is a required parameter. Avoid a 500 error when it's missing or empty. Instead generate a proper 400 bad request response.

Refs:

- https://datatracker.ietf.org/doc/html/rfc7033#section-4.2

P.S. Semi-related: Our default setup checks in `server` aren't compatible with `social` being enabled since they only check for 200 or 404 (neither of which are applicable when `social` is enabled and when querying `/.well-known/webfinger` without any parameterst):

https://github.com/nextcloud/server/blob/16812837157395c078a9689cd51530a6382e17d2/apps/settings/lib/SetupChecks/WellKnownUrls.php#L48

This PR doesn't impact those either way directly (since the prior 500 nor the 400 added via this PR would have ever passed). I'll try to take a closer look at those over in `server`. EDIT: nextcloud/server#49440

### DIFF
--- a/lib/WellKnown/WebfingerHandler.php
+++ b/lib/WellKnown/WebfingerHandler.php
@@ -224,7 +224,7 @@ class WebfingerHandler implements IHandler {
 		// on some setup (i.e. tests) the data are not available from IRequest
 		$requestUri = $request->getRequestUri();
 		if ($requestUri !== null) {
-			parse_str(parse_url($requestUri, PHP_URL_QUERY), $query);
+			parse_str(parse_url($requestUri, PHP_URL_QUERY) ?? '', $query);
 		}
 		
 		return $query['resource'] ?? '';
--- a/lib/WellKnown/WebfingerHandler.php
+++ b/lib/WellKnown/WebfingerHandler.php
@@ -223,7 +223,7 @@ class WebfingerHandler implements IHandler {
 		// work around to extract resource:
 		// on some setup (i.e. tests) the data are not available from IRequest
 		$requestUri = $request->getRequestUri();
-		if ($requestUri !== '') {
+		if ($requestUri !== null) {
 			parse_str(parse_url($requestUri, PHP_URL_QUERY), $query);
 		}
 		
--- a/lib/WellKnown/WebfingerHandler.php
+++ b/lib/WellKnown/WebfingerHandler.php
@@ -223,7 +223,7 @@ class WebfingerHandler implements IHandler {
 		// work around to extract resource:
 		// on some setup (i.e. tests) the data are not available from IRequest
 		$requestUri = $request->getRequestUri();
-		if ($requestUri !== null) {
+		if ($requestUri !== '') {
 			parse_str(parse_url($requestUri, PHP_URL_QUERY) ?? '', $query);
 		}
 		
--- a/lib/WellKnown/WebfingerHandler.php
+++ b/lib/WellKnown/WebfingerHandler.php
@@ -97,6 +97,12 @@ class WebfingerHandler implements IHandler {
 	 */
 	public function handleWebfinger(IRequestContext $context, ?IResponse $previousResponse): ?IResponse {
 		$subject = $this->getSubjectFromRequest($context->getHttpRequest());
+		
+		// the `resource` parameter is required
+		if ($subject === null || $subject === '') {
+			return new JrdResponse('', Http::STATUS_BAD_REQUEST);
+		}
+		
 		if (str_starts_with($subject, 'acct:')) {
 			$subject = substr($subject, 5);
 		}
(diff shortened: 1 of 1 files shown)


paulvt:
This did not help for me, the error I now get is:

```
[index] Error: parse_str(): Argument #1 ($string) must be of type string, null given in file '/var/www/html/custom_apps/social/lib/WellKnown/WebfingerHandler.php' line 228
	GET /index.php/.well-known/webfinger
	from [...] by -- at 21 Nov 2024, 21:03:53
```

joshtrichards:
@paulvt Oops. Try now. :)

paulvt:
Unfortunately still the same. And I see the static-psalm-analysis GitHub Action says that strings cannot be `null` too :wink:

joshtrichards:
Though I don't see how it should ever end up truly null. It should be `''`. Even psalm agrees...

oculos:
> This did not help for me, the error I now get is:
> 
> ```
> [index] Error: parse_str(): Argument #1 ($string) must be of type string, null given in file '/var/www/html/custom_apps/social/lib/WellKnown/WebfingerHandler.php' line 228
> 	GET /index.php/.well-known/webfinger
> 	from [...] by -- at 21 Nov 2024, 21:03:53
> ```

Same for me.

lib/WellKnown/WebfingerHandler.php
```diff
@@ -216,8 +222,11 @@ private function getSubjectFromRequest(IRequest $request): string {
 
 		// work around to extract resource:
 		// on some setup (i.e. tests) the data are not available from IRequest
-		parse_str(parse_url($request->getRequestUri(), PHP_URL_QUERY), $query);
-
+		$requestUri = $request->getRequestUri();
+		if ($requestUri !== null) {
+			parse_str(parse_url($requestUri, PHP_URL_QUERY), $query);
```
come-nc, line 227: parse_url will return null if there is no query.
See https://www.php.net/manual/fr/function.parse-url.php
parse_str will not accept null.
```suggestion
			parse_str(parse_url($requestUri, PHP_URL_QUERY) ?? '', $query);
```

paulvt: This seems to help, I can use curl to get `/.well-known/webfinger` now and I get back: `""` with HTTP 400.
This still makes the security check in the Administration Settings unhappy, but at least no error is logged anymore.

joshtrichards: You're right. Thanks, @come-nc.

@paulvt Setup check is being dealt with nextcloud/server#49440

karlitschek:
Superseded by #2048, which redoes the still-valid half of this patch on the current code (the `getParam('resource')` fallback landed separately in the meantime) and credits you as co-author. Thanks for the fix, @joshtrichards!

PR #2031, opened 2026-09-07
fix: repair the cron/queue failure paths and index the timeline hot path

karlitschek:
* Resolves: # 
* Target version: master

### Summary

A stability pass over the background jobs, the federation delivery queue, and the query layer — the things an operator feels on a busy instance. **Backend only; no `src/` changes.** Every fix ships with a regression test (verified to fail before the fix); the suite runs at **1478 tests / 4968 assertions**. App version bumped to **0.10.2** for the migration.

### Broken

| | |
|---|---|
| **One `occ social:account:delete` permanently killed all background caching.** `AccountService::manageDeletedActors()` used `self::TIME_RETENTION`, which was never defined; the resulting `Error` isn't an `Exception`, so it escaped the cron's `catch (Exception)` and stopped the whole run — avatars, actor refresh, hashtags, timeline sync never ran again. | Constants defined; `Cron\Cache` catches `Throwable` per step. |
| **Incoming activities were dropped behind a `200`.** `ImportService` swallowed every exception, so a DB error while storing a remote post lost it silently and the sender never retried. | Only understood-but-unprocessable cases are tolerated; real failures propagate → inbox returns 5xx → sender retries. |
| **The federation queue never gave up, pruned, or reaped.** A dead host was retried forever; a delivered request stayed as a permanent `STATUS_SUCCESS` row (unbounded table growth); a worker killed mid-delivery stranded a `RUNNING` row that was never retried (silent lost delivery). `setAsFailure` also left the in-memory status as `SUCCESS`. | Abandon after `MAX_TRIES`; delete on success; `Cron\Queue` reaps stale `RUNNING` rows back to standby; standby fetch bounded; status fixed. |

### Slow

- **Home timeline full-scanned the two largest tables on every request**, and hashtag trends and the "since" timeline scanned `social_stream.published_time` — none indexed. A migration adds `social_follow(actor_id_prim, accepted)`, `social_stream_dest(actor_id, type)`, `social_stream(published_time)`, and the **primary keys** `social_stream_dest` and `social_stream_tag` lacked (also unblocks Galera/`pxc_strict_mode`).
- **Hashtag trends hydrated every note in five windows into memory** → cron OOM on a busy instance. Bounded to the most recent notes per window. (A `COUNT(*) … GROUP BY` over `social_stream_tag` would remove the hydration entirely and is the better long-term fix; it needs integration-test coverage this suite lacks — noted in the code.)
- **`limit` was unclamped on every timeline endpoint** — `?limit=10000000` a memory fatal, negative invalid SQL. Clamped to `ProbeOptions::MAX_LIMIT` at both chokepoints.
- **A DM or reply always did one synchronous remote delivery inside the user's request** — `getPriorityRequest` compared a *status* against a *priority* constant. Fixed.
- **A remote outbox was re-fetched on every profile view** (now best-effort, so a slow remote no longer fails the page) and for **every** cached actor in one cron pass (now a bounded batch).
- **One anonymous GET to `accountFollowers`/`accountFollowing` could fan out to thousands of outbound fetches** — the limit bounded successes, not attempts. Limit clamped and the item walk bounded.

### Notes for review

- The migration adds an autoincrement primary key to two populated tables; on a large instance that is a one-time table rebuild.
- `KEY_PAIR_LIFESPAN` is defined for the dormant `blindKeyRotation()` (no scheduled caller); it is not wired up here — key rotation is a separate change.
- Independent of, and non-conflicting with, the open #2029 (frontend basics) and #2030 (security).

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -20,7 +20,7 @@ What it does:
 - 🌐 Federation: signed delivery of ActivityPub activities, with a delivery queue processed by cron
 
 This is a partial implementation of ActivityPub and of the Mastodon client API. It does not offer blocking, muting or reporting, approvable follow requests, polls, bookmarks, lists, or video and audio attachments.]]></description>
-	<version>0.10.1</version>
+	<version>0.10.2</version>
 	<licence>agpl</licence>
 	<author mail="benedikt.schaechner@web.de" homepage="https://benedikt.xn--schchner-2za.de">Benedikt Schächner</author>
 	<namespace>Social</namespace>
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "nextcloud/social",
 	"description": "Social app",
 	"license": "AGPL-3.0-or-later",
-	"version": "0.10.1",
+	"version": "0.10.2",
 	"minimum-stability": "stable",
 	"authors": [
 		{
--- a/lib/Controller/ApiController.php
+++ b/lib/Controller/ApiController.php
@@ -859,6 +859,12 @@ class ApiController extends Controller {
 	 * @return array
 	 */
 	private function fetchRemoteCollection(string $url, int $limit = 20): array {
+		// A remote page can carry far more entries than asked for, and each
+		// unresolved id below is a remote fetch, so both the limit and the number of
+		// entries walked are bounded — an anonymous caller must not be able to turn
+		// one request into thousands of outbound fetches.
+		$limit = max(1, min(ProbeOptions::MAX_LIMIT, $limit));
+
 		try {
 			$collectionData = $this->curlService->retrieveObject($url);
 		} catch (Exception $e) {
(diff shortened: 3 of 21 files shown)


karlitschek:
**CI note — the two red cells are an invalid PHP×Nextcloud pairing, not a defect in this PR.**

The only failures are `SQLite/PostgreSQL PHP 8.2 Nextcloud stable35`, and they fail at `occ maintenance:install` with *"This version of Nextcloud requires at least PHP 8.3"* — i.e. before any app code, migration, or test runs. The `phpunit-sqlite`/`phpunit-pgsql` workflows cross the app's declared PHP floor (8.1) against the **max** server branch (stable35), which now requires PHP 8.3.

Evidence it is not this change:
- Every valid cell passes, including the migration under a real database: **MySQL × stable35 on PHP 8.3, 8.4 and 8.5 all green**.
- #2030's `phpunit-sqlite` run succeeded ~50 min before this one, with no relevant `info.xml` change since (only the `<version>` bump) — consistent with `stable35` raising its PHP floor in the interim, which makes the 8.2×35 cell install-incompatible on any PR, not just this one.

I did not force it green, because the only ways to do so are to change the app's declared PHP support or to edit the org-template CI workflow — both are maintainer decisions, not something a stability PR should do as a side effect. Options for whoever picks this up: exclude the `php < 8.3 × stable35` cell in the workflow, or pin the sqlite/pgsql cell to the max PHP.

PR #2034, opened 2026-09-07
fix: brute-force protect the OAuth token endpoint and add alt text to images

karlitschek:
* Target version: master

### Security — brute-force protection on `/oauth/token`

`OAuthController::token` is `@PublicPage` and validates a `client_secret` and authorization `code`, but nothing throttled it — so those credentials could be guessed without limit. The endpoint now carries `@BruteForceProtection(action=socialOauthToken)` and throttles **every credential-failure response** (unknown client id, wrong secret, bad code); a successful exchange is not throttled.

Federation endpoints (inbox, shared inbox, actor, WebFinger) are **deliberately left unthrottled** — remote servers must reach them freely.

Two tests assert a wrong secret and an unknown client both return a throttled `401`.

### Accessibility

The avatar images in `TimelineEntry` and the preview in `MediaAttachment` had no `alt`. The avatars get an empty (decorative) `alt`; the media preview uses the attachment's own `description`, which is exactly the alt text a Mastodon client already supplies.

### Notes
- `js/` is rebuilt for the `src/` changes. No version bump — 0.10.2 (unreleased) already carries the frontend changes.
- (Branch name is a leftover from an earlier plan; the content is the OAuth/a11y hardening described here.)


karlitschek:

PR #2061, opened 2026-09-08
feat: render custom emoji from other instances

karlitschek:
Second of the four feature PRs (media types #2060 → **custom emoji** → polls → full-text search).

Remote `:shortcodes:` rendered as literal text because `Emoji` tags were dropped on import. Now:

- **Wire → model**: `ACore::extractEmojisFromTag()` keeps Emoji tags (https icons only — a `javascript:` icon URL is dropped, pinned by tests) as Mastodon `CustomEmoji` entries on statuses (`Stream`) and actors (`Person`).
- **Cache round trip**: both re-extract from the stored wire `source` on `importFromDatabase()` — no schema change needed.
- **API**: status entities gain the (previously missing) `emojis` field — contract test updated; account entities' `emojis` is now real instead of `[]`.
- **Web client**: post content renders shortcodes via the existing DOM-to-vnode pipeline (a new handler in `MessageContent`), display names via a new `DisplayName` render component (TimelinePost + UserEntry). Everything goes through `createElement` — never `innerHTML`; a hostile emoji URL can't inject markup (pinned by a test).
- Own custom emoji remain out of scope: `/api/v1/custom_emojis` stays `[]`, documented.

Tests: PHP (wire import incl. icon-less/non-https rejection, DB round trips for status and actor, contract pin) and JS (known shortcode → img, unknown stays text, hostile URL never becomes HTML). Suites green (1744 PHP / 575 JS). E2E on devel: a seeded remote note with an Emoji tag serves the emoji entry through the local-format export after a full DB round trip.


karlitschek:

PR #2068, opened 2026-09-08
feat: profile metadata fields

karlitschek:
Third of the batch (own polls #2065 → client essentials #2066 → **profile fields** → pinned posts → link previews). The name/value table under the bio — where Fediverse users put their website, pronouns or location. Until now `fields` was hardcoded to `[]`, so neither your own nor a remote account's fields ever showed.

### What it does

- **Model** — `Person` carries up to four name/value pairs, exported as `PropertyValue` entries in the actor's `attachment` and as `fields` (plus `source.fields` for the editor) in the account entity. Incomplete pairs are dropped, names capped at 255 and values at 500 characters.
- **Persistence** — remote fields ride along in the stored wire source (like emoji and remote counters, no schema change); local fields get their own `social_actor.fields` column, which **wins over the cache** so an actor refresh can't lose them. Migration + version bump to 0.11.6.
- **APIs** — `PATCH /api/v1/accounts/update_credentials` accepts `fields_attributes` (a list, or an object keyed by index as form-encoded clients send it); `PUT /api/v1/account/fields` serves the web client.
- **UI** — the profile page renders the table for every account and offers an editor (add/remove rows, max four) on your own. Remote values arrive as HTML, so only their **text** is rendered and only `http(s)` values become links — a `javascript:` value from a hostile server can never become a clickable link (covered by a test).

Not included: **link verification** — `verified_at` is always `null`.

### Tests

6 model specs (ActivityPub round-trip, cap-and-drop, non-PropertyValue attachments ignored, cache source, column-wins, account entity), 2 controller specs (`fields_attributes` as a list and keyed by index), 8 component specs, and a new integration test (`ActorsProfileFieldsTest`) that round-trips the column through MySQL/PostgreSQL/SQLite including unicode and quotes. All 8 new PHP tests and 7 of 8 new frontend tests were verified to fail without the implementation. Suites: 1782 PHP, 589 vitest, php-cs and stylelint clean.

### Verified on devel

```
PATCH fields_attributes  → fields: [Website, Pronouns] (the incomplete third entry dropped)
GET   verify_credentials → both fields returned, source.fields carries the raw values
GET   /@admin (AP)       → attachment: [{PropertyValue, Website, …}, {PropertyValue, Pronouns, …}]
occ social:cache:refresh → fields still present in both the entity and the actor document
PATCH fields_attributes:[] → fields: [], attachment key gone from the actor document
PUT   /api/v1/account/fields (no session) → 401, route registered
```


karlitschek:

PR #2084, opened 2026-09-08
fix: delete the Fediverse account with the Nextcloud one

karlitschek:
## The bug

The app listened for `UserUpdatedEvent` and for nothing else, so **deleting a Nextcloud user left their Social account fully alive**: the actor kept resolving over WebFinger, their posts stayed readable, and remote servers went on delivering to an inbox whose owner no longer existed.

An administrator deleting an account has no reason to suspect any of that, and no reason to know `occ social:account:delete` exists.

## The fix

None of the work was missing — only the trigger. `AccountService::deleteActor()` already tombstones the actor, drops what belongs to it, and federates a `Delete{Person}` so servers that cached it drop their copies too. It was simply never called.

`UserDeletedListener` calls it now. It looks the actor up by user id and deletes by the actor's *own* handle, since the two are not always the same. A user who never opened Social has nothing here and is not an error. A failure is logged rather than thrown — the Nextcloud user is already gone by that point, and throwing would leave the deletion half-applied with nobody able to finish it.

## Demonstrated on a live instance

The bug found its own illustration: a `cwtest` user I had deleted hours earlier still had a live actor sitting in `social_actor`.

```
aisha      aisha      deleted=-
deltest    deltest    deleted=2026-09-08 20:49:39   ← deleted with the fix in place
cwtest     cwtest     deleted=-                     ← deleted hours earlier, still live
admin      admin      deleted=-
```

Deleting the `deltest` user tombstoned the actor within the same second, and WebFinger went from answering to 404 while `admin` kept answering 200. (The orphaned `cwtest` actor has since been cleaned up by hand.)

## Verification

- 1915 unit tests.
- 6 listener cases: the happy path, a handle that differs from the user id, a user with no Social account, a failure during deletion, a failure during lookup, and an unrelated event.
- `ApplicationTest` now pins the registration, verified to fail when the `registerEventListener` line is removed — a listener nobody registers is exactly the shape of the original bug.

## One thing I noticed but did not change

A tombstoned actor's document returns HTTP 200 with an error body rather than `410 Gone`. That is pre-existing and identical for an actor that never existed, so it is not a regression — but `410` is the convention that tells remote servers to stop trying. Worth a small follow-up.


karlitschek:

PR #2085, opened 2026-09-08
fix: stop posting activities to ourselves over HTTP

karlitschek:
## The bug

`ActivityService` had no notion of "local" anywhere in it. A follower on *this* instance contributed our own shared inbox to the delivery list like any other, so every same-instance interaction was queued as an HTTP POST from the server to its own public address.

At best that's a round trip to be handed back something we wrote ourselves. At worst the server can't reach itself that way — a reverse proxy, split-horizon DNS or an SSRF guard is enough — and the delivery fails through fifteen attempts before being dropped, with real instances queued behind it the whole time.

## Why nothing local is lost

Recipients are written into `social_stream_dest` when the item is saved, and that is what puts a post in a local timeline. The HTTP delivery was never how anyone here saw it.

## Measured on a live instance

The development box can't reach itself and had been quietly dropping these. With `admin`'s only follower being the local self-follow that `checkStatusTableFollows()` creates:

```
=== WITHOUT the fix ===
before: 20 deliveries waiting
after:  21 deliveries waiting     ← one self-addressed delivery, doomed

=== WITH the fix ===
before: 21 deliveries waiting
after:  21 deliveries waiting     ← nothing queued
```

Both posts appear in the local timeline immediately, which is the point.

A note on method: my first attempt at this A/B showed "no change" in both arms and I nearly reported it as a result. The posts were 401ing on a stale app password and I was measuring nothing. The numbers above are from after I noticed.

## Verification

- 1911 unit tests.
- Two new `ActivityServiceTest` cases — a followers expansion that includes a local follower, and a direct target on this instance — both verified to fail with the filter removed.
- The test harness's queue mock returned a token even when handed zero targets, which the real `generateRequestQueue()` never does. Fixed, since the second case turns on exactly that.


karlitschek:

karlitschek:

PR #2089, opened 2026-09-08
feat: an unread notification badge that counts

karlitschek:
## The bug

The Notifications entry in the sidebar carried `counter: '0'` — hard-coded, never updated. The app has never told anyone that anything happened; you had to go and look.

## Why this is bigger than a one-line fix

Counting needs somewhere to count *from*, and the app had no notion of "read" at all — no column, no marker, nothing.

It keeps **markers** now, which is the shape the client API already asks for: a last-read position per timeline, so the same account read on a phone and in a browser agrees with itself.

| Route | Purpose |
|---|---|
| `GET /api/v1/markers` | read the positions (Mastodon-compatible) |
| `POST /api/v1/markers` | move them forward (`write` scope) |
| `GET /api/v1/notifications/unread_count` | the number the badge draws |

Third-party clients get all three.

## Decisions worth reviewing

**A marker only ever moves forward.** Two clients reading the same account each report their own position, and the one further behind must not un-read what the other has already seen.

**The count is capped at 99.** That's all a two-digit badge conveys, and a reader back from a fortnight away shouldn't pay for an exact number.

**The badge clears optimistically** when the notifications timeline renders — the reader is looking straight at what it was counting — and is re-read from the server if the marker won't save.

**Storage is a per-user config value, not a table.** One small row per person, written when a timeline is opened and read when a badge is drawn. Not worth a migration.

## A bug the unit tests could not have caught

I wrote the frontend against `entry.nid` and made my fixtures match. A `Notification` entity carries the row id in `id` and has **no `nid` at all**, so the marker was never sent — my tests passed against a shape I had invented. The end-to-end run is what caught it. Both the code and the fixtures now use what the server actually sends.

## Verification

- 1921 PHP tests, 742 vitest specs.
- 7 `MarkerService` cases including the backwards refusal, the version counter, and stored nonsense reading as unread rather than throwing.
- 5 controller cases including scope enforcement in both directions.
- 8 store cases and 3 for marking-on-view, the last verified to fail with the dispatch removed.
- End to end on a live instance: 1 unread → set the marker at the newest notification id → 0 unread, with a backwards write correctly refused.

## One incidental fix

The new store module declares `state` as a function rather than a shared object literal, unlike its four neighbours. My own test caught the leak: a second store built from the module inherited the first one's state. I have not converted the others — out of scope — but they have the same shape.


karlitschek:

PR #2090, opened 2026-09-08
feat: make the timeline usable without a mouse or without sight

karlitschek:
An accessibility audit of the frontend turned up, across 38 components:

| | |
|---|---|
| `role` attributes | **0** |
| live regions | **0** |
| `tabindex` | **1** (on a hidden input) |
| `:focus-visible` | **1**, against **5** suppressed outlines |
| `<h1>` | 1, only on the OAuth page |

This PR is the structural half. Alt text and live regions follow separately.

## The composer had no name

Its writing surface was an anonymous `contenteditable` div whose placeholder was painted by CSS (`content: attr(placeholder)`), so a screen reader got **nothing** from the app's primary input. It is now a named `textbox` with `aria-multiline`, the character count is read out in words once it is worth interrupting for, and the focus ring is visible rather than a one-pixel border tint.

## Six click handlers on things nobody could focus

The post timestamp (an `<a>` with no `href`, and the main way into a post), the attachment tiles, the "+N more" tile whose accessible name was "+", the search clear whose name was "✕", and the profile banner. The first four are buttons; the banner is decorative beside the "Change banner" button that already existed.

**Only images get a button wrapper.** Video and audio carry their own controls and nesting those inside a button is invalid — the new source guard for this exact class of bug is what caught me doing it while writing this PR.

## Like threw away your focus

It was two mutually exclusive buttons swapped by `v-if`, so pressing it unmounted the element under the reader's focus and dropped them to `<body>`. One button that relabels itself now, and both like and boost report `aria-pressed` rather than leaving the state to how an icon is filled.

## j/k moved a highlight, not the keyboard

It set a CSS class and scrolled. It moves real focus now, to entries that can hold it, and the new-posts pill sends focus to the first arrival instead of only scrolling the page out from under it.

## Also

Every timeline has an `h1` — visible where one was already shown, named for a screen reader everywhere else. A post is an `article` named after its author. The three unnamed dialogs have names. An unknown visibility renders a "not sure" icon instead of nothing, because silence there reads as *public*. The decorative "Following" indicator no longer takes focus to do nothing. Placeholder-only inputs have labels, and one untranslated `value="Follow"` is translated.

## Guards

Two new source checks: no click handler on an unfocusable element (with an escape hatch for `role`/`tabindex`/key handlers/`aria-hidden`), and no `outline: none` without a replacement.

## Verification

- 742 vitest specs, 1909 PHP tests, lint and stylelint clean.
- New behavioural tests for the article landmark, the timestamp button, the like button surviving a state change, boost reporting its state, `j`/`k` moving focus, entries being focusable, and every timeline's heading. The focus and landmark ones were each verified to fail with the change reverted.


karlitschek:

PR #2091, opened 2026-09-08
feat: let people describe the pictures they post

karlitschek:
## The gap

This client rendered other servers' alt text perfectly well and **could not produce any of its own**.

`createMedia` sent a FormData carrying only the file. The preview tile offered a Delete button and nothing else. Nobody was ever asked what a picture showed. So every image this instance federated went out undescribed, to every reader who cannot see it.

The server side was already complete — the media endpoints have accepted `description` since they were written, and the timeline renders one as `alt`. Only the way to write one was missing. (There is even dead CSS in `PreviewGridItem` for a description overlay that no longer exists, so there was one once.)

## What this adds

- a labelled description field on every attachment in the composer
- a marker on the tile of an attachment that has none
- a count in the send row: "1 attachment has no description"

**It is a nudge, not a gate.** Somebody posting a screenshot in a hurry is not the person to punish, and a client that refuses to post is a client people stop using. Nothing here can stop a post going out.

## Timing

Descriptions are sent once, as the post goes, rather than on every keystroke — a request per letter for something that only matters when the post exists. A description that won't save is reported and the post still goes.

## Verification

- 738 vitest specs, lint and stylelint clean.
- 5 composer cases (the field exists and its label points at it, the warning appears, it clears once written, the description is sent trimmed, nothing is sent for an undescribed attachment) and 2 store cases.
- **End to end on a live instance**: uploaded a real PNG, described it through `PUT /api/v1/media/{id}`, posted it, and read `description: 'a cat asleep on a keyboard'` back off the status in the timeline.


karlitschek:

PR #2092, opened 2026-09-08
feat: say what changed, and make the instance colours readable

karlitschek:
The three findings from the accessibility audit that aren't about keyboards. Companion to #2090 and #2091.

## Nothing was ever announced

New posts arrive, pages load, the end of a timeline is reached — and a reader who cannot see the screen was told none of it. There were **zero** live regions in the app. The only thing that ever spoke was a *failure*, because the toast library happens to use one.

The timeline has a polite live region now, carrying the one sentence worth saying: how many new posts are waiting, that posts are loading, or that there is nothing more. Not every page that loads while somebody scrolls.

## The instance colour was unreadable at some hues

It was generated at a fixed lightness, which cannot work. At 52%, blue is dark enough to carry white text and yellow is nowhere near — `hsl(60, 62%, 52%)` against white is about **1.5:1**, so whoever happened to be on a yellow instance got an unreadable badge.

The hue is still the instance's own; that's the entire point of the feature. The lightness is now whatever that hue needs for 4.5:1, which is a different number for every hue. The avatar ring also says the instance in words, since a coloured ring is nothing at all to a reader who cannot see it, or who cannot tell two hues apart.

The new test walks 400 hostnames, computes WCAG relative luminance for each distinct colour and asserts 4.5:1 — verified to fail against the old fixed lightness.

## The motion test I wrote was weaker than it looked

It was a whole-file substring check. It passed on a file with a dozen animations and one guard, on a guard whose selectors matched nothing, and on the words appearing in a **comment**.

It now strips comments, finds the `@media (prefers-reduced-motion: reduce)` blocks and asks what those blocks actually turn off, covers the shorthand properties and `scroll-behavior` it used to miss, and scans `css/` — which the `src`-rooted walker never looked at at all.

Verified by replacing a real guard in `Poll.vue` with a comment mentioning it: the old check passed, this one names the file and the count.

The underlying code was fine, for the record — all ten animated components are genuinely guarded. It was the test that wasn't holding them there.

## Verification

734 vitest specs, lint and stylelint clean. Each of the three changes was verified to fail with itself reverted.


karlitschek:

PR #2099, opened 2026-09-09
fix(media): serve a document by either of its copies, and rebuild moved links

karlitschek:
Pictures did not load in any timeline. Two independent causes, both at the point where a media link is resolved.

## A preview uuid never resolved

`DocumentService::getFromUuid()` looked its document up with `getByLocalCopy()`, which matches the `local_copy` column and nothing else. A `preview_url` carries the **resized** uuid, so `mediaOpen()` answered 404 for every preview link — and the timeline renders its grid from `preview_url`, so no picture appeared. There was no lookup by `resized_copy` anywhere in `CacheDocumentsRequest`.

`getByCopy()` now matches either column, which is what the two URL builders sitting next to each other — `getMediaUrl()` and `getResizedMediaUrl()` — always implied.

Measured on a live instance before the change: every `local_copy` uuid answered 200 and every `resized_copy` uuid answered 404, without exception.

## Stored links kept an address the instance no longer answers on

Media links are absolute, and they are written into `social_stream.attachments` when the attachment is created — so a row keeps whatever host was current then. That breaks an instance that changes domain, and it breaks quietly on one where cron writes a row under `overwrite.cli.url` while a web request would have used the request host.

Only the last segment, the document uuid, stays meaningful, so `MediaAttachment::asLocal()` rebuilds the link from it against the address in use now. Anything whose tail is not one of our uuids is somebody else's URL and is handed back untouched, so a remote `url`/`preview_url` is unaffected.

For that to take effect the database read path had to stop handing the stored arrays straight on: `Stream::importFromDatabase()` now hydrates them into `MediaAttachment`, so the one place that knows how a media link is shaped gets to build it. The wire-import path already did this via `Document::convertToMediaAttachment()`; only the database path bypassed the model.

Because the rebuild happens on read, existing rows heal themselves and no migration is needed.

## Testing

- 1987 PHPUnit (3 new) and 96 integration tests, all green; `php-cs-fixer` 0 of 503 files; `eslint` unchanged
- The regression guard lives in `tests/Integration/Db/MediaStorageTest.php`, where it uploads a real PNG and resolves **both** of the resulting copies through `DocumentService::getFromUuid()`. That is the assertion that fails on `master`.
- Verified against a live instance whose stored rows named two different old hosts (`localhost` and `localhost:8099`): after the change every attachment and preview link answers 200 with real image bytes, where the previews answered 404 before.
- `docs/API.md` now says which uuids the `/media/{uuid}` route accepts, and that it serves public documents only.

## Found while in here, deliberately not fixed

A cached **actor avatar** document is never marked public — `DocumentService::cacheLocalAvatarByUsername()` omits the `setPublic(true)` that `cacheLocalHeaderByUsername()` sets two functions below — so a remote actor's avatar 404s on the public media route. The frontend hides most of this by rendering local actors through the core avatar endpoint and only falling back to the Social URL for remote ones.

It is left out because the safe fix is not a one-liner: marking every cached `Image` public would expose media attached to followers-only and direct posts, so this needs a way to tell an actor icon from post media, which is a design decision rather than a patch.


karlitschek:

PR #2105, opened 2026-09-11
Federation parity: edits that apply, threads that resolve, media that loads, and signatures that name an actor

karlitschek:
Eleven of the thirteen federation gaps from the Mastodon-parity review. Six commits, 2594 PHP tests green, php-cs-fixer clean. **Two items are not in here** and are listed at the bottom.

### Edits never took effect on Mastodon *(large)*
An `Update` carried no `updated` and a freshly rewritten `published`. Mastodon reads an Update without `updated` as an implicit one — it refreshes a poll's counters and throws the new text away — so **every edit made here was invisible everywhere else**. `published` now stays put, `updated` is stamped and emitted, and a remote `updated` is stored so an edited remote post says when it was edited.

### Reply parents were never fetched *(large)*
A reply whose parent we were never delivered was stored orphaned and read "in reply to nothing". The parent is now queued for caching the way a boosted object already was, and its own parent after it, eight deep, after the inbox has answered. Reading a thread returned only direct replies, so conversations stopped one level down; the context walks the whole subtree now, bounded by depth and count and viewer-filtered as before.

### Media on followers-only posts *(medium)*
A picture on a followers-only post was a broken image on Mastodon: the media route served only what visibility marked public, and Mastodon fetches media unsigned — so the restriction protected nothing and broke every private post with an attachment. Media is served by its uuid, which is what it always was: an unguessable capability handed to the post's audience. A non-public attachment is cached by the reader's browser but not by a shared proxy.

### Language *(medium)*
Every local post federated as English and every remote post was shown as English — the language was a hardcoded default, the client's `language` was never read, and `contentMap` was neither emitted nor parsed. Mastodon's per-language filters key on this, so until now they filed everything wrong.

### Retry window *(medium)*
Delivery gave up after about twelve hours; Mastodon keeps trying for two days. A peer down for a day lost our posts for good. The schedule now runs to two days, quick at first and spread out after.

### Discoverable / indexable, and account migration *(medium-small / small)*
Neither flag was emitted, so Mastodon assumed both false: nobody here appeared in a directory, a suggestion or a search, with no way to ask to. Both are stored and emitted now, opt-in as on Mastodon, settable through `update_credentials`.

Moving in or out was impossible — `alsoKnownAs` had no setter, nothing built a `Move`, and a Mastodon export could not be read. Three commands close it: `social:account:alias`, `social:account:move` (verifies the target claims us, federates the Move, re-follows for local followers) and `social:account:import-follows`. A remote `movedTo` is parsed now too, so a moved profile no longer looks like a quiet one.

### Delete, NodeInfo, and the signature work *(small)*
A `Delete` went to followers and the instances the post reached, leaving it standing wherever it arrived by boost — boosters and repliers are addressed now. NodeInfo announced the theming title as `software.name`, so every server running this app looked like different software to crawlers; it is `nextcloud-social`, with the human name as `metadata.nodeName`, the schema's required fields, and real post/domain counts instead of hardcoded zeros.

**RFC 9421** signatures are verified alongside draft-cavage (structured fields, the §2.5 signature base, v1_5-sha256 and PSS-sha512, same key retrieval and exceptions, the RFC's own Appendix B vector in the tests). Outbound stays cavage, which nothing refuses.

And the hygiene item turned out to matter more than "hygiene": **a signature proved which server sent an activity, never which account**, so anyone with an account on a server could act as anybody else on it. An activity must now be signed by the actor it claims, unless the object carries a Linked Data signature — which is what makes a relayed or forwarded activity legitimate, and where Mastodon draws the same line. Update and Delete are likewise bound to the stored author rather than to a matching hostname.

Also: a bare string in `to`/`cc` no longer makes a public post a direct message, and hashtags link to a route that exists.

### Not in this PR
- **Per-user domain blocks** — started, produced nothing; needs its own round.
- **Quote posts (FEP-044f)** and **hashtag following** — deliberately left for a second round, since both cut across files this PR rewrites.
- **Relays** — I'd push back on building these at all for a Nextcloud instance; worth a decision rather than a default yes.

Docs: `docs/OCC-Commands.md` covers the three new commands (test-enforced). `docs/Architecture.md` still describes signature verification as cavage-only and should gain the 9421 path — that is in the follow-up docs pass along with the API docs for `language`, `updated` and the actor flags.


karlitschek:

PR #2126, opened 2026-09-12
The Mastodon compatibility report, items 3 to 7

karlitschek:
`docs/Mastodon-Compatibility.md` (#2123) ends with an order of work. This is everything on that list measured in **hours or days** — items 3 through 7, plus the endpoint gaps of §3.4. What is left is named at the bottom, with why.

### A privacy leak: `source` on every account entity (§3.6)

`source` carries `follow_requests_count` — how many people are waiting on an account's approval, which is nobody's business but that account's. The model emitted it from `exportAsLocal()`, so **every** Account entity this app produced carried it: somebody else's profile, a search result, a page of followers, and an anonymous read of any of them.

It is now built by the two credentials routes, which are the two that know they are answering the account itself. The model cannot leak it any more because it no longer has it — `exportSourceAsLocal()` is a separate method nothing else calls. Mastodon puts `source` on exactly those two routes too.

### A moderation bug: suspension federated nothing (§6)

Suspending a local account deleted everything it had posted here and stopped serving its actor, and every remote instance carried on holding a full copy. The takedown stopped at our own edge.

It now sends the same `Delete` the account's own deletion sends. Only for local accounts — a `Delete` we signed for somebody else's actor is not one any peer would act on — and a failure is logged rather than raised, because a moderator waiting on a delivery queue is a moderator who cannot moderate.

### Three profile fields accepted and dropped (§3.7)

A client's profile editor sends the whole form in one PATCH, so somebody changing their name, picture and bio together got a **200** and only the bio.

- `display_name` and `avatar` belong to the Nextcloud account, so they are written there and the actor cache refreshed. A backend that owns either (LDAP, SAML, anything provisioned elsewhere) makes the request a **422** — the profile looks unchanged either way, and only one of those two outcomes tells the user why.
- `bot` needed a column. It also decides the actor's *type*: an account marked automated that went on publishing `Person` would have told a client and a peer different things, so the flag and the type are set together.

The report suggested a 422 for all three. I implemented them instead: rejecting fields every client sends on every save would have turned "your avatar was ignored" into "you cannot edit your bio".

### The version string was hiding finished work (§3.5)

`3.5.0` was right when it was written. It is not now: editing with `/source` and `/history`, v2 filters, `/api/v2/instance` and `/notifications/unread_count` are all here, all 4.x, and a client that believes the string never asks. It says **`4.2.0`**.

What is still missing is announced rather than left to fail — `configuration.translation.enabled` is `false`, `urls` is empty (so a client knows there is no streaming), and a client that tries Web Push gets a 404 and falls back to polling, which is what it does against any server without a VAPID key.

### Three missing endpoints (§3.4)

- **`/api/v1/accounts/search`** — what every composer calls to complete a `@handle`. No client uses `/api/v2/search` for autocomplete, so mention completion failed in all of them. `resolve` goes and asks the handle's own server; `following=true` narrows it to people the viewer follows, which is what a client wants when completing a reply.
- **`favourited_by`** and **`reblogged_by`** — a tap on a favourite or boost count is no longer a dead end. Both resolve the status through the visibility filter first: who liked a post is as private as the post, so one the caller may not read is a 404 and no reaction of it is looked at.

### Testing

3746 unit tests green. Every change has a test; the two that could be wrong quietly were mutation-verified — remove the federated `Delete` and `testSuspendingALocalAccountTellsTheFediverse` fails; let it fire for remote accounts and `testSuspendingARemoteAccountFederatesNothing` fails. The entity contract test (`ApiContractTest`) pins both key sets and was updated deliberately: `source` out of Account, `poll` into Status as `null` rather than absent, which was another §3.6 item.

`docs/API.md` and `docs/Architecture.md` are updated in the same commit.

### What is left on that list, and why

| # | Work | Estimate | Why not here |
|---|---|---|---|
| 1 | The API at the domain root (§3.1) | days | Needs a webserver rewrite or root route registration plus setup guidance — a deployment decision as much as code |
| 2 | Per-user OAuth tokens (§3.2) | days | A new token table keyed to (client, user) and a migration off the single-token column |
| 8 | Web Push (§9.8) | weeks | |
| 9 | Authorized fetch inbound (§9.9) | weeks | |
| 10 | Stored identity, the Mastodon URL space, key import (§5) | months | A product decision, not a defect — only worth starting if in-place instance migration is an actual goal |

Items 1 and 2 are the two that stand between this and stock Mastodon clients; I can take them next.


karlitschek:

karlitschek:
Rebased onto master (which now carries #2123, #2124 and #2125) and **updated `docs/Mastodon-Compatibility.md` in the same branch**, so the report and the code it describes move together:

- §1 and the header say 0.16.0, and note what this wave changed for a reader who knew the document before.
- §3.4 no longer lists `accounts/search`, `favourited_by` and `reblogged_by` as missing; what is genuinely still absent (push, streaming, preferences, familiar_followers, peers, activity, v1 filters) stays.
- §3.5 describes the version string as it is — `4.2.0`, with translation and streaming announced as unavailable rather than left to fail.
- §3.6 marks the `source` leak and the absent `poll` key as fixed and keeps the two that remain (the empty `avatar` on `accounts/{id}`, and `redirect_uri` missing from `POST /api/v1/apps`).
- §3.7 and §6 record the profile fields and the federated suspension.
- §8 counts all four moderation findings as fixed.
- §9 keeps the full order of work with a **State** column, so the five done items stay visible next to the five open ones rather than disappearing.
- §10 keeps the one piece of documentation drift that is still true — the README's promise about third-party clients — and says why that kind is worse than the stale kind.

While in there I also replaced every `file.php:123` reference with the method name. Line numbers rot faster than anything else in a document like this, and four of them were already wrong.

karlitschek:

PR #2129, opened 2026-09-12
build(lint): move to ESLint 10 and @nextcloud/eslint-config 9

karlitschek:
Unblocks #2116 and #2122, which dependabot cannot land on its own.

### Why neither lint bump can merge today

`@nextcloud/eslint-config@9` requires `eslint >= 10` and the flat config format. #2116 bumps the config to 9 and dies at `npm ci`; #2122 bumps `eslint-plugin-n` to 18, which the version-8 config peer-pins to `^16`. Both fail with ERESOLVE before a single file is linted.

This migration is what unblocks them — and it makes both **moot**: version 9 bundles its own plugins, so `eslint-plugin-n`, `eslint-plugin-import` and `eslint-plugin-vue` stop being direct dependencies. Once this lands, the two bumps can be closed rather than merged.

### What changed

`.eslintrc.js` becomes `eslint.config.mjs`. The globals and three rule overrides carry across unchanged. Two additions the old setup did not need:

- `process` as a global — webpack substitutes it at build time and `src/store/index.js` reads it
- the vitest globals — the specs use `describe`/`it`/`expect` without importing them

The lockfile change is confined to the lint toolchain: 157 packages removed, 64 added, 35 version changes, **every one of them** in the eslint ecosystem. Nothing else moved — I checked the resolved-version map before and after.

### The 1200 errors that are not in this pull request

The new ruleset reports about 1200 further errors across 126 files: blank lines between component properties (523), import order (193), indentation (139), and so on. None of them find a defect. Adopting them belongs in its own change rather than riding along with a toolchain migration, so `eslint.config.mjs` turns them off in one block that names each rule and the count it would cost.

One deserves a specific note. `perfectionist/sort-imports` reorders 298 import lines and, doing so, **detaches the comments that explain why particular imports are there** — several are side-effect imports whose comment is the only thing saying so. That rule needs a human pass, not `--fix`.

### What the new ruleset found that was worth fixing

A 35-line diff across 20 files:

| Fix | Count |
| --- | --- |
| `catch (error)` bindings nobody read, now optional catch bindings | 16 |
| assignments overwritten before they were read | 3 |
| unused Vuex getter/action parameters | 3 |
| `v-bind` of a string literal, quoted prop, missing trailing comma | 4 |
| translated strings whose ellipsis could wrap alone, now non-breaking space | 6 |

⚠️ That last row changes six source strings (`Loading …`, `Search …`, `Following …`, `Unfollowing …`, `Searching …`), so **those six need retranslating**. They are short and common, so translation memory should re-match most of them.

`no-use-before-define` is configured to allow forward references to hoisted functions and module-level consts, which several files here use deliberately to read top-down from entry point to helper. `classes` stays on, where a forward reference is a real temporal-dead-zone error.

### Verification

`npm run lint` (0 errors), `npm run stylelint`, `npx vitest run` and `npm run build` all pass locally. `js/` is rebuilt in the same commit because the `src/` fixes change the bundle.


karlitschek:

PR #2130, opened 2026-09-12
Take the dependency bumps that can be taken

karlitschek:
Nine dependabot PRs, none of which could land on its own: each carries its own `package-lock.json`, so the first to merge makes conflicts of the other eight. Three were already failing CI outright. This is all of them that resolve, in one tree, with the fallout fixed — closing #2113, #2114, #2115, #2117, #2118, #2119 and #2120.

### In range, no code change

`@nextcloud/dialogs` 7.5.0 · `vue` and `@vue/compiler-sfc` 3.5.42 · `@nextcloud/babel-config` 1.3.0 · `webpack` 5.110.3 · `jsdom` 30.0.1.

### Two things that needed fixing

**`terser-webpack-plugin` is now a direct devDependency.** webpack 5.110 stopped hoisting it where `@nextcloud/webpack-vue-config` expects to find it, and the build died with `Cannot find module 'terser-webpack-plugin'` before compiling a byte. It was always a real dependency of this build — it was just being borrowed from somebody else's tree.

**vue-router 4.6.4 → 5.3.1 needed the vite toolchain to move with it** (`vite` 8, `@vitejs/plugin-vue` 6). v5 declares an optional peer on vite 7 or 8, and npm refuses to resolve it against the vite 6 the old test toolchain pinned — which is why #2114 has been sitting there conflicting.

One v5 behaviour change reaches this app: **an absent optional route param is `undefined` where v4 gave `''`**, so `/search` with no term now hands `Search.vue` nothing. Nothing breaks — the prop declares `default: ''` and `App.vue` already reads `to.params.term ?? ''` — but both halves are now asserted, in the router test and in the component test, so the next person to meet that `undefined` knows it is the router's answer and not a hole in the code.

### Two that are not here, and why

- **`node-polyfill-webpack-plugin` 4.1.0 (#2121)** — `@nextcloud/webpack-vue-config@6.3.2` peer-pins it to exactly `4.0.0`. There is nothing to fix on this side: it needs a release upstream. I left the PR open rather than closing it, since it becomes mergeable the day that lands.
- **`eslint-plugin-n` 18 (#2122)** and **`@nextcloud/eslint-config` 9 (#2116)** — both made moot by #2129, which took the whole ESLint 10 flat-config migration. The config bundles its own plugins, so neither is a direct dependency any more.

### Testing

`vitest` 1185 passing (one new), `npm run lint` clean, `npm run build` clean, and the PHP suite untouched and green. Bundles rebuilt and the app version bumped so browsers do not keep the old ones.


karlitschek:

PR #2132, opened 2026-09-12
Fix everything the performance survey still listed as open

karlitschek:
Everything `docs/Performance.md` still listed as open, fixed — and the document updated to match, so it keeps describing the code rather than the code's history.

### The last unbounded cron loop

`getNotCachedDocuments()` read *every* uncached document and fetched each one over HTTP in a single cron slot. An instance that had been offline for a day tried to fetch its whole backlog at once. Capped at `CACHE_BATCH`, which is what its two siblings (`SYNC_BATCH`, `STANDBY_BATCH`) already did.

### The unbounded read on the inbound path

An incoming `Delete` walked every post that had ever addressed the account, inline in the HTTP request the peer was waiting on. That is not merely slow — **the peer times out and re-sends, so the walk starts again from the beginning**, for ever, on exactly the accounts that have the most of it.

It now pages by **keyset, not offset**: the walk rewrites and sometimes deletes the posts behind those rows, so the set shrinks underneath an offset and every shift skips a row — which here means a post left addressed to an account that no longer exists. A bounded number are done inline and the rest goes to `Cron\ActorCleanup`, which re-queues itself while rows remain: the shape `Cron\DomainPurge` already uses.

### The N+1 behind the two hottest client routes

A relationship cost six queries — the two follow rows, the blocks and mutes, the note, the mute's expiry. So `accounts/relationships` for a page of forty cost **240 round trips**, and `accounts/search?following=true` paid up to eighty *per keystroke*.

Every one of those was already a lookup on (viewer, account), and `IN` takes a list: **five queries now, for any number of accounts**. The single-account route goes down the same path deliberately — two implementations of "what is the relationship between these two" is two places for the answer to differ, and the route that draws a follow button and the route that draws a list of them have to agree. The query count is asserted and mutation-verified, because an N+1 comes back as one innocent lookup inside a loop.

### Three more

- **The whole hashtag table, every cron run.** `manageHashtags()` read it to find the few rows that have fallen out of every window; `getWithAnyTrend()` reads only the rows that claim a trend, which is the only thing it can change.
- **The follower list with no cursor.** `/api/v1/current/followers` hydrated every follower into memory. Bounded at 500, and `docs/API.md` now says so and points at the Mastodon route that pages.
- **A post that exists and is in nobody's timeline.** `StreamRequest::save()` wrote the post, then its recipients, then its tags, outside any transaction — and `StreamDestRequest::create()` logged its failure and carried on. Since the recipient rows are what put a post in a timeline, a partial write was **permanent and silent**. One transaction now, and the recipient insert raises, so a delivery can be retried into a clean state.

### Testing

3761 unit tests green. Two behaviours were mutation-verified: remove the inline bound and the `Delete` walks everything (the test fails cleanly rather than hanging — I rewrote it when the first version only hung); ignore the keyset cursor and the paging test fails. The N+1 test fails the moment the batch call goes back inside a loop.

The transaction is proven by an **integration** test rather than a unit one: a transaction is a database behaviour, and there is nothing to assert about it without a database. It runs in the PHPUnit MySQL/PostgreSQL/SQLite jobs, not in the unit suite I can run locally.

### What is left in the survey

The eighteen read paths still deduplicating over ~120 columns (notifications named as where to start), `StreamActionService::saveAction()` and `ModerationRequest::save()` wanting the transaction treatment, and the schema items.


karlitschek:
Pushed a second commit carrying the changes into the rest of `docs/`, so each document describes what it is responsible for rather than the survey carrying all of it:

- **`OCC-Commands.md`** — `Cron\ActorCleanup` in the background-jobs table, which is where an admin looks to find out what a job on their instance is. A queued job whose argument is an actor id is otherwise a mystery, so the row says what it is for.
- **`Architecture.md`** — the same job in the registration table, and the directory tree corrected: it listed three jobs when there are five, two of them queued with an argument rather than registered in `info.xml`.
- **`Technical-Debt.md`** — its "what has been done" table now records that the unbounded work, the N+1 and the missing transaction closed here, and points at `Performance.md`, which it already names as the owner of query behaviour.

`DocumentationTest` is green, which covers the checkable parts (routes, schema rows, command options); the prose above is the part no test can check, which is why it moves in the same commit as the code.

PR #2133, opened 2026-09-12
The full inventory of what a Mastodon replacement still needs

karlitschek:
§9 of `docs/Mastodon-Compatibility.md` was a ten-row "suggested order of work", five rows of which were already done. It is now **the whole list** — thirty items in five tiers, every one re-checked in the code today rather than carried over, each with a rough size and what it actually fixes.

### The shape of what is left

**Tier 1 — two blockers.** The API is not served at the domain root, and an app row holds exactly one token. Until both land, no stock Mastodon client can reach *any* of the surface below, and a second person signing in through Elk or Phanpy signs the first one out.

**Tier 2 — what a client shows and cannot get.** Web Push, `/preferences`, custom emoji, `familiar_followers`, `instance/peers` and `/activity`, the v1 filter routes, the `saved_searches` stub, and streaming.

**Tier 3 — what a peer would notice.** No authorized fetch inbound (signature verification runs on inbox POSTs only, so a followers-only object cannot be served to an authorized remote reader — it fails closed, nothing leaks), no `Add`/`Remove` for pins, an empty `mediaType`, the WebFinger profile link pointing at the Nextcloud profile, and no emoji reactions either way.

**Tier 4 — the admin and moderation surface.** Registration, approval queue and invites; warnings and strikes, so the ladder does not jump from silence to suspend; an account browser and a takedown button in the web UI; graded domain blocks; IP and email-domain blocks; a moderator role that is not "full Nextcloud admin"; metrics; the remaining `tootctl` equivalents.

**Tier 5 — the takeover.** Thirty rows down to six, and they are one project: stored actor identity comes first because everything else depends on it, then the Mastodon URL space, key import, the id rename, the importers, reconciliation and a cutover check.

### Two things the audit found, fixed here rather than listed

- **A local bot account was served as a `Person`.** `parseActorsSelectSql()` set the type on every read, *after* `importFromDatabase()` had derived it from the new `bot` column — so the flag said `true` to a client and the actor document said `Person` to every peer it federated with. That is the two of them disagreeing about one fact, and I introduced it in #2126. The type follows the flag now, with a test pinning it.
- **`POST /api/v1/apps` omitted `redirect_uri` and `vapid_key`.** Mastodon's Application entity always carries both, and a client that decodes the answer into a typed struct fails without them.

3758 tests green.


karlitschek:
Added a second commit: the list now has **a file of its own**, `docs/Mastodon-Roadmap.md`.

It was §9 of the compatibility report, which is the wrong home for it — that document answers *how close is this, and in which sense*, and somebody working through the gaps wants the backlog without the assessment wrapped around it. §9 keeps the shape (a five-row table of what each tier is and why it sits there) and points at the file for the detail, so there is one place for the detail to be wrong in rather than two.

Linked from the README beside the compatibility report, and added to the `DocumentationTest` case that keeps these files from being lost — it exists, it states the version it was verified against, and the README still links it. Mutation-verified: remove the file and the test fails.

`docs/` now holds seven documents, each with one job:

| File | Answers |
|---|---|
| `API.md` | which routes exist and what they do |
| `Architecture.md` | how the app is put together |
| `OCC-Commands.md` | what an admin can run |
| `Mastodon-Compatibility.md` | how close this is to Mastodon, in three senses |
| **`Mastodon-Roadmap.md`** | **what is still left to build, in tiers** |
| `Performance.md` | what costs what, and what is still slow |
| `Technical-Debt.md` | what is old, borrowed or load-bearing |

PR #2152, opened 2026-09-13
fix(social): stop follow-nextcloud button overlapping welcome banner

jospoortvliet:
## Summary

The "Follow Nextcloud on mastodon.xyz" button in the welcome banner overlaps surrounding text (see screenshot below). The button used `float: inline-end` inside `.follow-nextcloud`, but nothing cleared the float, so the container's height collapsed and the button overlapped neighboring content instead of sitting below the paragraph.

Fix: use flexbox (`display: flex; flex-direction: column; align-items: flex-end`) instead of float, so the button lays out in normal flow, right-aligned below the paragraph.

## AI disclosure

This change (root-cause analysis and the CSS fix) were drafted with Claude Code and reviewed by me before submission, per the [Nextcloud AI Contribution Policy](https://github.com/nextcloud/.github/blob/master/AI_POLICY.md).

## Test plan

- [ ] Open the Social app as a first-run user (or reset `firstrun` flag) so the welcome banner shows
- [ ] Confirm the follow button no longer overlaps the paragraph/banner text
- [ ] Check both LTR and RTL layouts


karlitschek:
thanks for the PR. The complete interface was replaced in the meantime

PR #2210, opened 2026-09-16
perf: what breaks first at a million users, measured and fixed

karlitschek:
Everything in the scale assessment, implemented and measured. **Based on `feat/peertube-parity` (#2209)**, which it expects to land first.

## How this was measured

`occ social:benchmark` seeded devel to **402,725 posts, 805,212 recipient rows and 2,044 actors**. Where a before/after appears below, both were run against the same machine in the same pass — devel's speed drifts by a factor of four under load, so numbers taken minutes apart are not comparable and only the ratio is honest.

The seeder was the first thing built, and it had to be rewritten to be useful: it seeded *through the model layer* at about 400 rows a second, which makes ten million posts a seven-hour wait, so nobody had ever seeded enough to find out what the queries cost. Bulk `INSERT`s do 6,000 a second.

## The three things that broke first

**The home timeline sorted everything you had ever been sent.** `EXPLAIN` said `Using temporary; Using filesort`: the planner drove from the viewer's follows, fetched every recipient row every followed account had ever produced, joined each to `social_stream` to find out when its post was published, and sorted the lot to keep twenty. One page load cost Σ(all posts of everyone you follow), and a client asks for one every thirty seconds.

The sort key is on the recipient row now — `social_stream_dest.nid`, safe to denormalise because a nid never changes after the row exists — with `(actor_id, type, nid)` behind it. **1,661 ms → 50 ms**, against a public timeline of 4.2 ms as the control. Index-only, no temporary table. The per-viewer filters (blocks, mutes, hidden boosts) were what forced the join, so they now run against the twenty rows the page chose rather than across everything the viewer follows. The old query stays as the fallback for an instance whose backfill has not finished.

**A cron job loaded every local account into one PHP array, every twelve minutes** — `getAll()`, no limit, a dozen queries and an avatar read behind each row, and no deadline. At a million accounts that array does not fit in memory; at a hundred thousand it does and the work is hours in a job that fires every twelve minutes, silently. It pages and resumes now. Separately, the three counters lived only in a JSON blob, so moving one meant recomputing all of them with four aggregate queries — **on every post written and every follow accepted**. They are columns that get incremented; the JSON is still what is read, and the cron's walk is what reconciles drift.

**Delivery had a ceiling of about a thousand an hour, and ten at worst.** `Cron\Queue` reads 200 rows every twelve minutes, sequentially, 30-second timeout each, inside a 300-second budget — so ten unresponsive peers fill the whole pass. `occ social:worker` is the same delivery in a loop that does not stop, and **several may run at once**: claiming a row was already atomic, so this scales by adding processes rather than by rewriting anything. The circuit breaker moved to the distributed cache with per-host backoff (it used to be emptied at the start of every pass, so a dead peer was rediscovered every twelve minutes). And the inbox no longer walks a whole thread inline while holding an FPM worker.

## The rest

- **`media_kind`** column and index, replacing `attachments LIKE '%"type":"video"%'` — which no index could serve and which was applied after the join.
- **Search is bounded** by `search_window_days` (a year). A leading wildcard can never use an index; this bounds the scan rather than pretending otherwise.
- **`304` on the polls.** The home timeline head and the unread count carry an `ETag` built from the newest id the viewer can see — one index-only probe, far less than the page it stands in for. Media carries the stored file's own tag; a timeline is forty to sixty pictures a screen, each of which was a full Nextcloud boot.
- **The first screenful is in the page.** Without it the first screen was a staircase: fetch the bundle, mount, *then* ask the server.
- **A covering index on `social_follow`** so the read that starts every home page never touches a row — 158 of its ~250 bytes are text ids that query does not want.
- **Two tables were leaking** on every post deletion (`social_watch`, `social_team_post`), and the actor lookup is memoised per request.
- **`Admin.md` gained a "Running a large instance" section**: install notify_push (a tenfold cut in poll volume for one app install), run workers, watch the cron actually finish.

## One mistake worth reading

The first version of the guard on the fast path asked the database *"is any recipient row still missing its nid?"* — a question with no index to answer it, and therefore a full scan of the largest table in the app: **427 ms on 800,000 rows, on every request**. It would have made the home timeline slower than the thing it replaced. The migration writes a flag instead. It was caught by measuring rather than by reading, which is the whole argument for having built the harness first.

## Two things deliberately not done

**Moving `source`/`cache` out of `social_stream`.** My own assessment overstated this. Both are read on the *local* path — a post's `source` carries a poll's options, an actor's carries its header, fields and emojis — so it is a side-table split rather than a narrowing. And the home-timeline fix removed most of its urgency: those five kilobytes are now read for the twenty rows of a page instead of for everything the viewer follows.

**Materialised home feeds.** Not needed yet, by the decision procedure the assessment set out: the cheap version gave 33×, and the read is now bounded by the viewer's own follow graph rather than by the instance's history. The `nid` column is what the materialised version would build on, so nothing here is wasted if it turns out to be needed.

**`X-Sendfile` / presigned media URLs** are also absent, and cannot be done here: `ISimpleFile` exposes no local path, and on object storage there is none. That needs a storage-level answer, not an application-level one.

## Verified

5,691 PHP tests and psalm clean at **every one of the twelve commits**, not just the tip; 2,345 JS tests, eslint, stylelint, typecheck and build clean. Both migrations were run on the seeded instance: the recipient backfill took 86 s over 805,212 rows, the media-kind backfill 8 s over 402,725 posts.


karlitschek:
Two more commits, one of which is a regression I introduced and then caught by reading a measurement rather than a test.

**A narrowed home timeline came back empty.** Moving the per-viewer filters off the page query and onto the twenty rows it chose was right for blocks and mutes, which drop a few posts — and wrong for a media narrowing, which drops nearly all of them. A media question is about the *post*, and the page query reads only the recipient rows, which carry no such column. On the seeded instance 143 of 402,725 posts carry media, so Photos and Videos read the newest sixty ids, threw all sixty away, and returned nothing while the pictures sat a few thousand rows further down.

Those timelines take the join path now, which has the predicate in the query. They are read far less often than the home timeline, and with `media_kind` indexed they are **faster than the fast path was**: 3.7 ms and the right answer, against 45.7 ms and no answer.

Every test passed while this was broken, because none of them had a timeline where most posts carry no media. `HomeTimelineFastPathTest` now pins all four refusal rules against the real method — it was mutation-checked, and fails on each of them if the condition is removed.

Latest numbers on the seeded instance (402,725 posts, 805,212 recipient rows):

| | |
|---|---|
| home timeline | 40.4 ms |
| public timeline (control) | 2.9 ms |
| home + media | 3.7 ms |
| `ETag` probe (what a 304 costs) | 0.27 ms |
| follows read (covering index) | 0.46 ms |

The `ETag` probe is the one that validates the `304` design: a validator has to cost far less than the page it stands in for, and at 0.27 ms against 40 ms it does.

PR #2221, opened 2026-09-17
fix(navigation): the sidebar calls the blocking page what the page calls itself

karlitschek:
## What is wrong

The app calls one page two things:

| Where | What it says |
|---|---|
| the sidebar entry | **Blocked and muted accounts** |
| the top of the page it opens | **Blocking** |

## How it got that way

`ac85fdff` (*feat(blocking): one card per decision, instead of four runs of text*, 17 Sept)
renamed the page from **Blocked and muted accounts** to **Blocking** and gave it a card per
decision — Blocked, Muted, Hidden servers, Filtered words. **The sidebar entry that leads
there was not renamed with it.** It has been the odd one out since that afternoon.

## Which name wins, and why

The page's, and not only because it is the newer one. That entry leads to a page holding
four lists and named two of them: a server you have hidden and a word you would rather not
read are neither an account nor something you muted. The docs had already moved — they say
*Blocking → Filtered words* in three places — so the menu was the last thing still using
the old words.

## What this changes

- `Navigation.vue`: the entry is **Blocking**.
- `docs/User-Guide.md` (two places) and `docs/API.md` (one), which still spelled out the
  old name.
- `Navigation.test.js`, which asserted the old label in five places and would otherwise
  have gone on asserting it.

Nothing else moves: the route is still `/blocked`, and the page, its cards and its API are
untouched.

## Verified

2430 JS tests, 5821 PHP tests, `tsc` typecheck, eslint, stylelint — all clean. Bundle
rebuilt in the release commit.

Version 0.26.3 rather than 0.26.2 because #2220 is open and claims that one.


karlitschek:
Superseded by #2222, which is now on master: it carries the same rename (sidebar entry, docs and the tests that named it), plus the avatar cache-busting fix and the filtered notifications moved onto that page. Nothing here is left to merge — rebasing this branch would produce an empty diff — so closing rather than leaving it conflicting. Reopen if I have read it wrong.

PR #2274, opened 2026-09-20
Stable29

karlitschek:
* Resolves: # 
* Target version: master


karlitschek:
Closing without merging: `stable29` is a maintenance branch at 0.7.0, 1405 commits behind master (now 0.26.59). The 37 commits it carries that master lacks are Transifex translation syncs, an npm audit fix and a terser-webpack-plugin declaration — merging it into master would mean resolving a 180-file conflict between 0.7.0-era code and everything since, with no upside. The branch stays as-is for stable29 maintenance.

